PR #50, opened 2014-07-24
adding request object to parameter of rename function, onFileUploadData ...

rbg246:
Hi,

This is a fantastic module, a great drop-in replacement for multipart.

I've made a few minor changes to enable access to the request parameter in the renaming function, the onUploadStart event and onUploadComplete Event to provide more data for the handling of the file upload and post-processing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules/
 *.DS_Store
 temp/
+uploads/
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ The following are the options that can be passed to Multer.
 * `limits`
 * `includeEmptyFields`
 * `inMemory`
-* `rename(fieldname, filename)`
+* `rename(fieldname, filename, reqParams)`
 * `onFileUploadStart(file)`
 * `onFileUploadData(file, data)`
 * `onFileUploadComplete(file)`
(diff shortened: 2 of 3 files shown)


joangit:
hi, I have the same question like ' euglv' , i want to change the dest,  how to do it? 
example: images in '/path/images' ,for other files in '/path/files', just diffrent request can change the save path,thanks.

rbg246:
The changes in the pull request do this as my experience has been that the non-file fields are parsed by onUploadStart.

So you can use my pull request to do what you want.

joangit:
hi: 
it's my coffee code
app.use multer
  #dest:path.join(__dirname,config.systemPath),
  rename: (fieldname,filename) ->
    if fieldname is 'image'
      return 'backgroundImage'
    return filename
  onFileUploadStart:(file)->
    if file.fieldname in ['image','zipFile']
      console.log file.fieldname
      file.path = path.join(__dirname,config.systemPath,file.name)
    else
      console.log file.fieldname
      file.path = path.join(__dirname,config.filePath,file.name)
    console.log file
    console.log file.path

i use multer ,onFileUploadStart function ,    if file.fieldname in ['image','zipFile'],do something of new file.path,others do something of other path, it's success,but the file is not upload,  if i set 'dest', all files upload 'dest' path, the 'file.path' is unused, i'm lost, ... thank you

jpfluger:
Hey @rbg246. I've pushed a number of changes, some related to this pull request... as well as others. The [README](https://github.com/expressjs/multer) has been updated as well. A new function has been added to change the upload directory. Also `rename` and the onFileUpload* functions have both `req` and `res` passed into it. Get the userid from the session object, passed in by req or res. Then you can build the upload path or rename the file, as appropriate. Hopefully this gets you what you need.

rbg246:
Great i shall look into it.

Thanks for all your work!

PR #69, opened 2014-10-17
ability to see the file extension before rename

isidro-amv:


LinusU:
This is solved differently in `1.0.0`

PR #81, opened 2014-12-03
Include request, response objects in arguments to onFileUploadStart etc.

juliangoacher:
This PR is similar to #50 but is more generally useful because it passes both the request and response objects to the onFileUploadStart, onFileUploadData, onFileUploadComplete functions.

The main motivation for doing this is to allow progressive responses to be sent to the client (via either http or socket.io) as the upload is received. Access to the request object allows identifying info about the request to be read; access to the response object allows progress data to be sent to the client over http.

The new arguments are additional to the existing file + data args, so the change is backwards compatible.

### DIFF
--- a/index.js
+++ b/index.js
@@ -98,7 +98,7 @@ module.exports = function(options) {
 
         // trigger "file upload start" event
         if (options.onFileUploadStart) {
-          var proceed = options.onFileUploadStart(file);
+          var proceed = options.onFileUploadStart(file, req, res);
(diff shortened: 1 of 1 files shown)


bigmonkeyboy:
+1 for this PR, having req and res available would be very useful (especially on upload complete)

lucasmotta:
+1 for this too

jpfluger:
This has been added. See the [README](https://github.com/expressjs/multer) for more info. I'll contact the npm maintainers so they can get it updated there too.

PR #85, opened 2014-12-18
Cleaner extension split and fix for redundant ext null check

vukicevic:
Nothing major. I changed the .slice(-1)[0] to .pop(), it seems cleaner. 

Also, the variable ext is checked === null, which is never the case as it is always set to be some string. I've changed the conditional check so null will be returned if ext was empty. The conditional check can be removed completely to retain the current behavior: empty string when no extension is present.

### DIFF
--- a/index.js
+++ b/index.js
@@ -77,8 +77,7 @@ module.exports = function(options) {
         // defines is processing a new file
         fileCount++;
 
-        if (filename.indexOf('.') > 0) { ext = '.' + filename.split('.').slice(-1)[0]; }
-        else { ext = ''; }
+        ext = filename.indexOf('.') > 0 ? '.' + filename.split('.').pop() : '';
 
         newFilename = rename(fieldname, filename.replace(ext, '')) + ext;
         newFilePath = path.join(dest, newFilename);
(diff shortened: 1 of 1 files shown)


jpfluger:
Actually, `pop()` removes the ending extension completely from filename, which then causes line 88 to be wrong `originalname: filename,`.  `slice()` will not remove the ending extension.

I agree that the `ext === null` check is never evaluated in line 93: (Nice eyes, by the way :+1:)

```javascript
extension: (ext === null) ? null : ext.replace('.', ''),
```

But your suggested solution of...

```javascript
extension: ext === '' ? null : ext.replace('.', ''),
```

...is not fully backwards compatible and might break users installations. I think the following would be safer b/c it's backwards compatible, since the empty quotes were already the default:

```javascript
extension: ext === '' ? '' : ext.replace('.', ''),
```

jpfluger:
Per my comments above, I went ahead and committed a [minor fix](https://github.com/expressjs/multer/commit/aac1c9b9073d0990561d99fe1f95c305d3c8683a), which @vukicevic correctly identified didn't do anything.

vukicevic:
Good points. Thanks for taking the time to review the pull request. I'll try to be more careful in future requests.

PR #97, opened 2015-02-02
Added an option to remove file extension after rename;

hotab:
Did not add any tests, but seems to work correctly;

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ module.exports = function(options) {
   options = options || {};
   options.includeEmptyFields = options.includeEmptyFields || false;
   options.inMemory = options.inMemory || false;
+  options.removeExtensionAfterRename = options.removeExtensionAfterRename || false;
 
   // if the destination directory does not exist then assign uploads to the operating system's temporary directory
   var dest;
@@ -80,7 +81,13 @@ module.exports = function(options) {
         if (filename.indexOf('.') > 0) { ext = '.' + filename.split('.').slice(-1)[0]; }
         else { ext = ''; }
 
-        newFilename = rename(fieldname, filename.replace(ext, '')) + ext;
+        if(options.removeExtensionAfterRename) {
+            newFilename = rename(fieldname, filename.replace(ext, ''));
+        }
+        else {
+            newFilename = rename(fieldname, filename.replace(ext, ''))+ext;
+        }
+
         newFilePath = path.join(dest, newFilename);
 
         var file = {


jpfluger:
The way you have it will introduce a minor bug on line 96, which is 

```javascript
extension: (ext === null) ? null : ext.replace('.', ''),
```

because the `ext` value will have been the original value.

Curiously, why do you desire to get rid of the extension, if there was one there originally?  What purpose does it serve? I ask because I used to do something similar and ran into problems trying to debug files. The extension is usually mapped to an app that can open it.

hotab:
The user might, for example, transmit some characters impossible for the filesystem to store, or is it handled internally somewhere? Also, won't it open some sort of Directory Traversal when used with, for example, nginx, or am I missing something? (https://en.wikipedia.org/wiki/Directory_traversal_attack). 
And some use cases do not require the extension (for example when you use easyimage module, it will use Imagemagick, which works without it).

jpfluger:
Re. [filename limitations] (http://en.wikipedia.org/wiki/Filename#Comparison%5Fof%5Ffile%5Fname%5Flimitations). If the filesystem running multer is Unix-based, I wouldn't worry about it, since only null and forward-slash (`/`) are disallowed. If the filesystem is NTFS, then there is a chance for exceptions, which could be handled by renaming the file to an Uuid using the rename function. In addition, Windows has [additional problems](http://www.howtogeek.com/127154/how-hackers-can-disguise-malicious-programs-with-fake-file-extensions/) above-and-beyond what we're discussing. 

The chances that the extension would not be compatible between systems is very small. Is this a real-world issue for you?  You'll need to convince me or one of the other maintainers (eg @hacksparrow, @Fishrock123) that this code is needed here or simply use your changes in a fork of multer.

Re. directory traversal attacks. Not related to this. Multer is middleware for an app-process based on node.js. The app-process will run separately from Nginx. Nginx will call the app-process from within its configs, usually based on IP:Port number. I have [examples](https://github.com/jpfluger/examples) for setting-up node to be used with nginx. Nginx can be configured to pass everything on-to the app-process or server some-things itself, like static content. Check that Nginx disallows directory-traversal, esp for static stuff. And in the app-process, if you map routes to paths (one-to-one), this has the effect of disallowing directory-traversal. 

Please note if an uploaded file is saved to a directory accessible by url, then the files in that directory could be downloaded. Check your access controls to ensure only authorized users can access those files.

Re. easyimage/imagemagick. Yes, I've used ImageMagick to identify unknown files for me in the past. That's because I had gotten rid of the extension in the first place! I did not know about EasyImage. That could be useful somewhere. Thanks.

hotab:
Thanks for the explanation, and example configs for running node with nvm and good example of nginx usage - will definitely look into that. I am closing this request (feature is not required).

SargTeX:
Hello,
we would really need this feature for several projects. We're creating a file object in the database to store all the metadata we need. The filename results in being the id of the file object.
We do this for several reasons:
 1. To make sure each file name is unique and is equal to another, already stored, file.
 2. To remain the filename upon updating the file itself.
 3. To store additional metadata (e.g. creator notes aso) and make it searchable.
 4. For some (weak) security reasons.

Storing the filename with an extension (such as 54fc7357a9f1ff8416f2dafb.PDF) would result in some conflicts upon updating the file, as well as require additionally storing the extension (or the filename on hard disk) beside the "virtual filename", containing a file extension, created by the user.

Are you sure you don't want to add this to multer? I really appreciate the commit.

jpfluger:
When we come out with the next major version, this might be the time to rethink file naming. Instead of passing in `fieldname, filename, req, res`, we pass an object for the `filename`. The default values of this object would include `directory, filename and extension` and you can now override them in one place. We could include the `encodetype` and `mimetype` as well but nothing happens if someone did change these values, but I say include them anyways.  Then pass back the file object and processing moves forward.  @hacksparrow thoughts?

hacksparrow:
@jpfluger yea, we could have this ability and the change in parameter type in the upcoming next major version.

LinusU:
This is solved differently in `1.0.0`

PR #102, opened 2015-03-07
Cloud Object Storage Support

dustin-H:
Hi,
I created support for pkgcloud/pkgcloud which supports some Cloud Object Storages like Openstack or Amazon S3. My functionality does not affect the default functionality of multer.

May you like to add this to multer.
Sorry my code beautifier inserted some spaces.

Best regards
Dustin Hoffner


jpfluger:
Interesting. Useful. Code looks like it pipes directly to the cloud stream and not to disk. Do you have mocha tests for this?  Looks like the [pkgcloud](https://github.com/pkgcloud/pkgcloud/tree/master/test) repo has its own tests. Possible to copy examples from there within multer test?  The readme would need updating. Probably part of a major version release. @hacksparrow thoughts?

dustin-H:
Yes it just creates a writeStream and pipes it to the cloud storage. I wanted to write some mocha tests but I had no idea how to do this without a server. 
Thanks for the good idea to have a look on pkgcloud tests. Now I know. I'm going to write some the next days.
I will also add some documentation to the readme and improve some of my code.

iamsebastian:
Could be really useful for handling media uploads at our API. 
:+1:

dustin-H:
Now. Sorry for the long time since last commit. I just wrote tests for testing pkgcloud upload. However they require a running Cloud Object Storage so without the tests will fail.

LinusU:
Hi, I just saw your work here after posting #123. I think that it's quite relevant and that it would be a really nice way to solve this.

You could then create a package called `multer-pkgcloud` that implemented that storage engine :)

dustin-H:
Hi LinusU, thanks!
I already created a npm module called [multer-pkgcloud](https://www.npmjs.com/package/multer-pkgcloud) if you meant that? 
I'm using this for my customers webapp, which goes productive the next days, so if there should occur  some problems I am going to fix them quickly.

LinusU:
Ahh, sorry, that wasn't what I meant. In #123 I described my idea to implement an api for storage modules. If this was implemented with that you wouldn't need to maintain a separate fork, instead you would have a repo with just the storage engine. The usage could look something like this:

```javascript
var mutler = require('mutler')
var pkgcloud = requrie('mutler-pkgcloud')

var storage = pkgcloud({ ... })
app.use(mutler({ storage: storage }))
```

This is only ideas thought and nothing is implemented. Hopefully I'll get more time to devote to this during the weekend :)

dustin-H:
Nice, sorry.. Now I understood. I like this idea and would help to implement it, however it only makes sense if the maintainer of multer would support it.

LinusU:
I'm now a maintainer and it seems like there is positive support form the others, hopefully I could whip up a pull request during the weekend and than we can have further discussions with the current maintainers. But I got the impression that they are very friendly and open to changes for the better!

dustin-H:
Ahh, nice to know. (I should have looked into the repo info ;) ) I look forward to see your work and keep watching.

LinusU:
`1.0.0` is now released, I've opened an issue in your repository to help you get it compatible with the new version.

PR #106, opened 2015-03-20
Updated to busboy 0.2.7

xulien:
Contains major bugfix => multipart: add workaround for joyent/node@2efe4ab


xulien:
missclick

hacksparrow:
The latest version is actually 0.2.9.

PR #111, opened 2015-03-27
REQ/RES as options for onFileSizeLimit to send a direct respond (e.g. 413 Entity too large)

derMani:
onFileSizeLimit, onFilesLimit, onFieldsLimit, onPartsLimit now can be utilized to send a direct respond to the requesting site. For example: onFileSizeLimit could be called, but it was difficult to send a HTTP 413 (Request Entity Too Large) directly back to the client.

There are also a few whitespaces and semicolon fixes (complained by JSLint).

### DIFF
--- a/index.js
+++ b/index.js
@@ -106,7 +106,7 @@ module.exports = function(options) {
         if (options.onFileUploadStart) {
           var proceed = options.onFileUploadStart(file, req, res);
           // if the onFileUploadStart handler returned null, it means we should proceed further, discard the file!
-          if (proceed == false) {
+          if (proceed === false) {
             fileCount--;
             return fileStream.resume();
           }
@@ -121,9 +121,9 @@ module.exports = function(options) {
         }
 
         fileStream.on('data', function(data) {
-          if (data) { 
+          if (data) {
             if (options.inMemory) bufs.push(data);
-            file.size += data.length; 
+            file.size += data.length;
           }
           // trigger "file data" event
           if (options.onFileUploadData) { options.onFileUploadData(file, data, req, res); }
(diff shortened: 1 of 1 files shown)


pfirpfel:
I think, in your approach you break the middle ware concept. If you perform a res.send() in one of the limit functions, multer will just continue and write to the response object after it was send leading to error.

As an example (HTTP 413), I recommend the following approach:
Delete the truncated (incomplete) file in onFileSizeLimit(). Then, in onParseEnd(), you can look if there is a file, where the truncated property is set to true. If this is the case, the FileSizeLimit was exceeded. You can now propagate the error, by handing it over to the next function as described [here](http://expressjs.com/guide/error-handling.html).

LinusU:
Have a look in #128 

How would this api suit you?

```javascript
app.use(function (err, req, res, next) {
  
  if (err.code === 'LIMIT_FILE_SIZE') {
    res.status(413).end('You have uploaded too large files!')
    return
  }

  // Your normal error handling
})
```

This change would make it very easy to integrate with your existing error handling. Most projects have an error handler like this that sends back nice pages, you just have to check `err.code` and set the status appropriately.

atorkhov:
How is it possible to terminate parsing when limit is reached?

LinusU:
@atorkhov I'm pretty sure thats how it works now, at least that is how it will work when #131 lands :)

LinusU:
The api I posted before is now live with multer `1.0.0`, the parsing will be terminated as soon as the limit is reached.

derMani:
That should suit my needs well! Good work

LinusU:
Thank you :+1:

PR #126, opened 2015-05-01
[WIP] Storage API

LinusU:
*Don't pull yet*

I took a stab at implementing what my thoughts where on the Storage API without actually breaking backwards compatibility.

@dustin-H Would you be open to try and implement `pkgcloud` on top of this? Or I could do it for you and maybe you could help me try it out :)

So this is the basic specification for a plugin:

```javascript
// The implementation could be a class or just a plain object
// This example uses a class
function PkgcloudStorage (opts) {
  this.client = opts.client
  this.getDestination = opts.destination
}

// The class or object needs to expose one function: handleFile(req, file, cb)
PkgcloudStorage.prototype.handleFile = function handleFile (req, file, cb) {
  var that = this

  that.getDestination(req, file, function (err, dest) {
    if (err) return cb(err)

    var params = {
      container: dest.container,
      remote: dest.remote,
      contentType: file.mimetype
    }
    var outStream = that.client.upload(params)

    file.stream.pipe(outStream)
    outStream.on('error', cb)
    outStream.on('success', function (info) {
      cb(null, {
        size: info.size,
        container: dest.container,
        remote: dest.remote
      })
    })
  })
}

// The module should expose one function that returns the storage instance.
// It should take one parameter: opts
module.exports = function (opts) {
  return new PkgcloudStorage(opts)
}
```

It would be used something like this:

```javascript
var mutler = require('mutler')
var pkgcloudStorage = require('mutler-storage-pkgcloud')

function destination (req, file, cb) { cb(null, { container: 'a', remote: 'b.txt' }) }
var storage = pkgcloudStorage({ client: _, destination: destination })

mutler({ storage: storage })
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,11 +1,13 @@
-var os = require('os');
-var fs = require('fs');
 var path = require('path');
-var crypto = require('crypto');
 var Busboy = require('busboy');
 var mkdirp = require('mkdirp');
 var is = require('type-is');
 var qs = require('qs');
+
+var extend = require('xtend')
+
+var diskStorage = require('./storage/disk')
+var memoryStorage = require('./storage/memory')
 
 module.exports = function(options) {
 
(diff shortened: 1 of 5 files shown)


LinusU:
Shit, so I managed to break all the tests with my final little tweak :laughing: will fix asap

LinusU:
Okay, so I actually want to make a slight change to the api, I'll update my first post shortly

LinusU:
I removed `file.req` and added `req` as the first parameter to the `handleFile` call.

I don't think we should expose `res` because the only purpose it serves is to send a response back to the client.

dustin-H:
Hi, sure, I'm going to implement and test it this weekend.

jpfluger:
:+1:

dustin-H:
Hi, I implemented and documented the pkgcloud plugin. It is tested on my local device with an external openstack object cloud storage. I tried different files/sizes and multiple files. All worked! :)  

I created an extra repository and npm package for it: 

- https://github.com/dustin-H/multer-storage-pkgcloud 
- https://www.npmjs.com/package/multer-storage-pkgcloud

I haven't wrote any tests so far. In my opinion they do not make that much sense since travis does not support any cloud object storage to test. See http://docs.travis-ci.com/user/database-setup/#Services-(data-stores%2C-messaging-brokers%2C-etc)-in-the-Travis-CI-Environment .

I have not that much experience with travis, so may there is a cloud object storage support I haven't seen. 

What do you think?

LinusU:
Nice! I left you a message about a bug in the commit, otherwise it looks good! :+1:

jpfluger:
Great job @dustin-H.

Re. module integration w/in multer. To be clear, is the going plan to add pkgcloud to multer/storage?  With idea others get added, like Mongo's GridFS? Or we keeping packages separate? Thoughts?

LinusU:
I really think that we should keep the other ones in there separate packages. Memory and disk is going to be quite common *and they don't depend on any external packages` so I think that they should stay.

If we start adding other ones we need to update multer every time there is a change in e.g. `pkgcloud` or `mongodb`. It would also mean that we take it upon ourselves to maintain every storage engine which I don't think is a good idea either...

dustin-H:
Thanks! :) I'll have a look at the bug this evening. In my opinion it wouldn't be that big problem to integrate the pkgcloud storage into multer since it has absolutely no dependencies. Even not pkgcloud because the client object needed gets passed in options. However I think keeping storages generally seperate would be the better solution, cause integration testing with too many different storage types gets very hard.

LinusU:
Closing in favour of #131

PR #140, opened 2015-06-04
[TASK] pass file extension to rename.

misgina:
The smallest  change possible. 

I understand there is similar pull request, but this is more backward compatible.

by having the file extension, it would have been possible to handle the name conflict manually and generate the filenames accordingly.


LinusU:
I'm sorry if I'm not up to date but what issue does this solve? Do you have an example or bug number?

LinusU:
This is superseded by `1.0.0`

PR #143, opened 2015-06-17
index.js update with Initial Validation

Tito47875:
Added initial file validation.
Set options.useValidation = true
options.validationString = "%PDF" for example, and all files starting with "%PDF" will be rejected

Regards,


LinusU:
This will need to be implemented again as all code has been changed for `1.0.0`, discussion in #155.

PR #165, opened 2015-07-20
package: node 0.8.x support

LinusU:
I don't see any reason not to support Node.js 0.8.x, even if it's quite old. If the Travis tests pass without issues I think we should merge this one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: node_js
 node_js:
+  - "0.8"
   - "0.10"
   - "0.12"
+before_install:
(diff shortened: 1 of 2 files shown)


LinusU:
Hmm, so none of the tests work because `stream.PassThrough` isn't available on `0.8.x`. Maybe we should just drop this...

jpfluger:
That was a fun try!

LinusU:
I don't really see any benefit of supporting `0.8.x`, closing this one...

OussamaRomdhane:
Yeah I closed it because the Travis tests didn't pass. Most modules in the express framework support 0.8.x so I thought this one should too out of coherence. My bad :smile:

Issue #171, opened 2015-07-22
Body parser, req.body is instantiated on null prototype


Maximization:
In my project I'm using multer together with Joi to validate incoming body requests. I've noticed this was giving me an error of undefined while other non multipart type requests going through body-parser worked just fine. I managed to track the error to the fact that Joi uses hasOwnProperty on req.body while multer instantiates req.body on null as Object.create(null), this can be found [here](https://github.com/expressjs/multer/blob/master/lib/make-middleware.js#L28). Therefore none of the global object functions are available to use on req.body.

It might be nice to instantiate req.body as an object literal `{}` or `Object.create(Object.prototype)` to have those methods available on the parsed body for further use down the middleware chain. Body parser already does this as you can see [here](https://github.com/expressjs/body-parser/blob/master/lib/types/json.js#L96).

Let me know what you think and if you agree. I can submit a quick PR if you want to.

LinusU:
Hmm, the reason I did this was specifically not to have the native prototype on it. The problem with letting the prototype be on it is that any user is free to at will overwrite those properties. E.g. a user could send a key called `hasOwnProperty` and thus overwrite that function.

Here are some good reading on the topic: [Blog post by Guillermo Rauch](http://www.devthought.com/2012/01/18/an-object-is-not-a-hash/), [Article on 2ality](http://www.2ality.com/2012/01/objects-as-maps.html)

I'm feeling that the correct solution here would be to patch Joi not to use `hasOwnProperty` from the provided user input, but instead using it from the native object prototype.

```javascript
function hasOwnProperty (obj, prop) {
  return Object.prototype.hasOwnProperty.call(obj, prop)
}
```

I'm very confident that they would accept the patch since it's made for validating user input, and user input can always contain strange stuff. In the worst case scenario here, an end user could trick the system into executing code (although quite unlikely, it's definitely possible).

LinusU:
See hapijs/joi#693

LinusU:
Hmm, I actually believe that the problem is not wit joi but with something else. Are you maybe using [express-joi](https://github.com/petreboy14/express-joi)?

Maximization:
Good article, I agree with you on the topic that we should never trust user input. I like the proposed solution of using inherited properties on the native object. Thanks for spending the time trying to fix it on their end.

[...]

PR #187, opened 2015-07-31
Update make-middleware.js. Object.create(null) => Object.create(Object.prototype)

alexkruegger:
The problem is when you are using Object.create(null), JS created absolutely null object without inherited properties even from Object.prototype. So for Body object it is critical as we can't use (for example) hasOwnProperty method as below:

            if ( req.body.hasOwnProperty(field) && Model.schema.path(field) ) {
                model[ field ] = req.body[ field ];
            }

More info here: http://stackoverflow.com/questions/15518328/creating-js-object-with-object-createnull

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -25,7 +25,7 @@ function makeMiddleware (setup) {
       default: throw new Error('Unknown file strategy: ' + fileStrategy)
     }
 
-    req.body = Object.create(null)
+    req.body = Object.create(Object.prototype)
 
     var busboy = new Busboy({ headers: req.headers, limits: limits })
     var isDone = false


LinusU:
Using `hasOwnProperty` on user supplied values is an antipattern. What would happen to your code if the user uploaded a field called `hasOwnProperty`?

Here are some good reading on the topic: [Blog post by Guillermo Rauch](http://www.devthought.com/2012/01/18/an-object-is-not-a-hash/), [Article on 2ality](http://www.2ality.com/2012/01/objects-as-maps.html)

I would recommend patching the tools you are using to use the following function:

```javascript
function hasOwnProperty (obj, prop) {
  return Object.prototype.hasOwnProperty.call(obj, prop)
}
```

Please see #171 for a very similar discussion

alexkruegger:
OK, I see. The proposed solution is nice. Thank you!

LinusU:
No problem :+1:

Issue #194, opened 2015-08-07
Partially uploaded files are not removed when fileSize kicks in


qqilihq:
I'm using the `fileSize` property to set a maximum upload size limit. In case, the uploaded file size exceeds the specified `fileSize` limit, the transfer is interrupted (as expected), but the partially uploaded file still exists on disk. 

I feel, that such files should be removed automatically (also concerning the fact, that there seems to be no way of retrieving the path of partially uploaded files within my callback methods).

LinusU:
I think I might have squashed that bug last night, or maybe not... Which version of multer are you using?

qqilihq:
Hey Linus, thank you for your feedback; I've just updated to 1.0.3, and now the file is deleted in case it exceeds the `fileSize` limit. 

However: The solution which you were describing in #186 for custom error handling does not work any longer with 1.0.3:

    var upload = multer({ limits: { fileSize: 1024 * 1024 * 1024 } });
    
    app.post('/upload', upload.single('file'), function (req, res) {
      // will now also be called in case the limit was reached
      res.send({ result: 'ok' });
    });
    
    app.use(function (err, req, res, next) {
      if (err.code === 'LIMIT_FILE_SIZE') {
        // will not be reached any more :(
      }
    });

As I was also relying on the custom error handling: Any solution on getting that fixed?

PokerCurve:
I am having the same issue as @qqilihq where the err.code is not being set to 'LIMIT_FILE_SIZE' when the limit file size is exceeded in 1.0.3.

A workaround is to perform the following check

    if(typeof req.file.filename == "undefined")  {
       // There was an error with the upload		
    }

qqilihq:
@PokerCurve Thank you for the suggestion, however in that case I cannot distinguish between the cases where (1) the limit was exceeded or (2) no file was attached at all. And I would really like to do that for better error messaging.

LinusU:
This should absolutely work, I'm sorry to have broke it.

This is very related to nodejs/io.js#2323 which there is a proposed fix for. Unfortunately I'm on vacation on a remote island with very bad internet connection and limited time to devote to this.

-----

Sorry, I just found this message sitting in the textbox, I forgot to hit "comment" :confused:

qqilihq:
Hi Linus, I've fixed the issue and sent you a pull request. Enjoy your vacation.

LinusU:
@qqilihq Could you try #205 and see if it fixes it for you?

[...]

PR #200, opened 2015-08-12
Fix and test for issue #194

qqilihq:
Fix for error handling for `fileSize` limit; in case diskStorage was used, route was called when limit was hit, instead of error.

### DIFF
--- a/test/express-integration.js
+++ b/test/express-integration.js
@@ -15,21 +15,21 @@ var port = 34279
 
 describe('Express Integration', function () {
   var uploadDir
-  before(function(done) {
+  before(function (done) {
     temp.mkdir(function(err, dir) {
       uploadDir = dir
-      done();
-    });
+      done(err)
+    })
   })
-  after(function() {
-    rimraf.sync(uploadDir)
+  after(function (done) {
+    rimraf(uploadDir, done)
   })
   it('should work with express error handling', function (done) {
     var app = express()
     var limits = { fileSize: 200 }
-    var upload = multer({ 
-      limits: limits, 
-      storage: multer.diskStorage({destination:uploadDir}) 
+    var upload = multer({
+      limits: limits,
+      storage: multer.diskStorage({destination: uploadDir})
     })
     var form = new FormData()
 
--- a/test/express-integration.js
+++ b/test/express-integration.js
@@ -16,7 +16,7 @@ var port = 34279
 describe('Express Integration', function () {
   var uploadDir
   before(function (done) {
-    temp.mkdir(function(err, dir) {
+    temp.mkdir(function (err, dir) {
       uploadDir = dir
       done(err)
     })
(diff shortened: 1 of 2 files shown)


LinusU:
Would you mind fixing the following style issues so that the travis build will run :+1: 

```text
standard: Use JavaScript Standard Style (https://github.com/feross/standard)
  test/express-integration.js:18:17: Missing space before function parentheses.
  test/express-integration.js:19:15: Expected error to be handled.
  test/express-integration.js:19:23: Missing space before function parentheses.
  test/express-integration.js:21:13: Extra semicolon.
  test/express-integration.js:22:7: Extra semicolon.
  test/express-integration.js:24:16: Missing space before function parentheses.
  test/express-integration.js:30:26: Trailing spaces not allowed.
  test/express-integration.js:31:22: Trailing spaces not allowed.
  test/express-integration.js:32:47: Missing space before value for key "destination".
  test/express-integration.js:32:59: Trailing spaces not allowed.
```

test/express-integration.js
```diff
@@ -8,14 +8,29 @@ var util = require('./_util')
...
 describe('Express Integration', function () {
+  var uploadDir
+  before(function(done) {
+    temp.mkdir(function(err, dir) {
+      uploadDir = dir
+      done();
+    });
+  })
+  after(function() {
+    rimraf.sync(uploadDir)
```
LinusU: Not that important but I would prefer:

```javascript
after(function (done) {
  rimraf(uploadDir, done)
```

qqilihq:
Sorry, fixed now.

LinusU:
No problem, I'm trying to figure out why this works. I remember that I had the `abort` in an upper scope, such as you have now, but moved it into a tighter scope. I don't remember why thought. Would you mind explaining the code?

Also, maybe the check for aborting should be moved to `indicateDone` instead of in `multer.on('finish')`...

qqilihq:
I moved the checking of the aborting flag, as suggested.

My problem was, that in case I was using the disk storage, and the limit was hit, the express route handler was getting called, instead of the error handler. That's why I introduced the `!aborting` check within the indicateDone (in case, the upload is canceled because of file size limit, the `done(uploadError)` callback (line 54) will be triggered.

Not sure why the CI test is failing now, works locally for me?

LinusU:
I restarted the CI job, it seems like it worked this time :confused:

qqilihq:
Hi Linus, not sure about the reason for the fail above. Running it locally several times gave no failure at all.

LinusU:
This should be fixed by #205 now :)

qqilihq:
Hi Linus, thank you, however I'm still having that problem. As soon as I use the `diskStorage`, the error event is not called when exceeding `fileSizeLimit`. In case of using the default storage, everything is fine. Test case demonstrating the issue is my first commit above.

Best,
Philipp

Issue #163, opened 2015-07-20
Dynamic fieldnames


polonel:
Before 1.0.0 I was using the rename function heavily to rename files based on a dynamic fieldname. I do not see an easy way to go about doing this anymore, and with requiring and fieldname on the single() call, am I going to have to rethink the way I was handling dynamic fieldnames?

sgulgas:
Looks like you'll have to make a very simple diskStorage Object

You can find the info on this section of the front page - they give a good example of it! : https://github.com/expressjs/multer#diskstorage

LinusU:
So there is basically two parts in this problem.

**1) Renaming files based on fieldname**

This is easily accomplished:

```javascript
var storage = multer.diskStorage({
  destination: '/tmp/uploads',
  filename: function (req, file, cb) {
    cb(null, file.fieldname)
  }
})
```

**2) Allowing any fieldnames**

This is a bit more complicated. We actively opted to not allow this as of now because of the security problems that arose with the previous behaviour. I'm afraid that allowing this will encourage bad behaviour. Not necessarily by us but I can easily see people asking others for help being told to use that functionality without understanding the ramifications of that.

The problem with accepting any fieldnames is that you as a user of this library must take extra care to actually process all the fields uploaded.

E.g.
```javascript
var upload = multer({ ... })

app.post('/upload-picture', upload, function (req, res, next) {
  req.files['pic'].forEach(function (file) {
    // Add the pictures to the users photo album
  })
})

// A malicious user could now upload any files and stick them into the
// temporary directory by making a POST request with an unexpected fieldname.
//
// e.g. POST a huge file with the fieldname blah to /upload-picture
```

Would you mind explaining your use case so that we can discuss how this best could be resolved?

polonel:
I have two sections that dynamic fieldnames are used.
1. The entire form is dynamic and is generated in real-time depending on a *metadata* schema defined by the end user. This results in the field names pulling from database and in some cases appends a "_X" (where X is a num) for multiple fields of the same type / name.

For example: One file upload could for an avatar image with another being for a zip of resources tied to this object. I wouldn't know in the codebase what the end user has named these fields in the schema. In the past It didn't matter as much because Multer was just accepted all fields and I would match the fields to the Schema after the post was complete.

LinusU:
Is the schema uploaded as well with the request? Is the schema saved on the server and then being transmitted to the client, who will then upload the files?

[...]

PR #131, opened 2015-05-01
[WIP] Breaking changes

LinusU:
I have now updated the README.md to a state which I think is quite good. I've tried to scale back and simplify as much as possible.

My intentions is to try and have a discussion here on how the api for the next breaking change should look like.

What I've changed:

- Removed `includeEmptyFields`, this will follow the W3C JSON form spec
- Removed `putSingleFilesInArray`, as discussed before this is now always true
- Removed `rename`, `changeDest` and `inMemory` in favour of `storage`, discussion in #126, #123 
- Removed `onFileUploadStart`, `onFileUploadData` and `onFileUploadEnd` in favour of `storage`
- Made some slight changes to the `File` properties, discussion in #130 
- Removed the rest of the `on*` in favour of `next(err)`, discussion in #128 

Todo:

- [x] It's no longer possible to skip files, maybe have a `filesFilter` option that's a function `(req, file, cb)`
- [x] Document the API for people who want to implement their own Storage Engine

This isn't implemented yet and I'm not saying that this is they way it's going to be, it's an early draft to a proposal. Comments very welcome :)

[Read the new README.md](https://github.com/expressjs/multer/blob/4c4cb61d417cadbcbf5ebfb7e2b1463c9af6e757/README.md)

PR #201, opened 2015-08-12
Add .matches()

kylev:
This allows people to use dynamic forms that may generate fields with
regular names, such as clicking "Add row" and getting a new field named
`description_47` and `file_upload_47`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,16 @@ Example:
 ]
 ```
 
+#### `.matches(regex[, maxCount])
+
+Accept a mix of files with names matching a regular expression. Optionally error
+out if more than `maxCount` files are uploaded. An object with arrays of files
+will be stored in `req.files`.
+
+For example, you might use `/^pdf_upload_\d+$/` if you have a dynamic form that
+needs to assocate numbered fields with files in a manner not guaranteed by
+browser field/upload ordering.
+
 ### `storage`
 
 #### `DiskStorage`
--- a/index.js
+++ b/index.js
@@ -66,6 +66,37 @@ Multer.prototype.fields = function (fields) {
   return this._makeMiddleware(fields, 'OBJECT')
 }
 
+Multer.prototype.matches = function (regex, maxFiles) {
+  function setup () {
+    var filesLeft = Infinity
+    if (typeof maxFiles === 'number') {
+      filesLeft = maxFiles
+    }
+
+    function matchingCountFilter (req, file, cb) {
+      if (filesLeft <= 0) {
+        return cb(makeError('LIMIT_UNEXPECTED_FILE', file.fieldname))
+      }
+
+      if (regex.test(file.fieldname)) {
+        filesLeft -= 1
+        return cb(null, true)
+      } else {
+        return cb(makeError('LIMIT_UNEXPECTED_FILE', file.fieldname))
+      }
+    }
+
+    return {
+      limits: this.limits,
+      storage: this.storage,
+      fileStrategy: 'OBJECT',
+      fileFilter: matchingCountFilter
+    }
+  }
+
+  return makeMiddleware(setup.bind(this))
+}
+
 function multer (options) {
   if (options === undefined) {
     return new Multer({})
--- a/test/expected-files.js
+++ b/test/expected-files.js
@@ -94,4 +94,47 @@ describe('Expected files', function () {
       done()
     })
   })
+
+  it('should accept files with matching fieldnames', function (done) {
+    var form = new FormData()
+    var parser = upload.matches(/^butme_\d$/, 5)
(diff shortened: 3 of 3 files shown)


gabeio:
Multer accepts all fields which aren't files. Files are the only ones which you should specify. You also might want to look at #163 it might help. As the entire point behind only allow fields from X,Y,Z or an array of files under a name which might be what you are looking at is that there was the capability of a DoS attack for more info please look at #131

kylev:
Yup, I think I understand the files and DDoS protection. I'm just trying to deal with a (legacy) page that uses a fairly common technique of generating sequential form names when you click on "Add Row". For example, the page loads with two fields: `description_0` (input=text) and `file_upload_0` (input=file). When the user clicks on "Add Row" the client side JS generates and inserts a new section into the DOM adding `description_1` and `file_1` (again, text and file respectively).

This is a pretty common client-side technique for supporting multiple file uploads with strongly associated data (the description for x has to go with the file for x). I understand that multer doesn't deal with the text fields; they're there to satisfy the specification of the app ("Allow me to upload up to 4 files with associated descriptions.").

I could do something like this (intentionally lame for clarity, I wouldn't type it out like this):

```javascript
multer().fields([
  { name: 'file_upload_0', maxCount: 1},
  { name: 'file_upload_1', maxCount: 1},
  { name: 'file_upload_2', maxCount: 1},
  { name: 'file_upload_3', maxCount: 1}
])
```

The equivalent with my proposed patch is:

```javascript
multer().matches(/^file_upload_\d$/, 4)
```

Both are identically DDoS-safe.

Does that make more sense of what I'm hoping to accomplish?

Beyond this there are quite a few client-side idioms for which there are going to be regular (in the regex sense) file upload fields. "For every input foo_x there will be a file upload bar_x," where "x" is a sequence number, UUID, randomize character string, etc.

gabeio:
I can see how the maxFiles would help DoS protect it but it is possible to say .array("file_upload") and on the page basically say name="file_upload[0]" name="file_upload[1]"...

kylev:
I'm not sure I understand... it would be great if this test case worked, but it fails.

```js
it('should accept arrays with bracketed names', function (done) {
  var form = new FormData()
  var parser = upload.array('ordered')

  form.append('ordered[0]', util.file('small0.dat'))
  form.append('ordered[1]', util.file('small1.dat'))

  util.submitForm(parser, form, function (err, req) {
    assert.ifError(err)

    assert.equal(req.files.length, 2)
    done()
  })
})
```

gabeio:
Okay I think I misunderstood I took a look at the existing tests: https://github.com/expressjs/multer/blob/master/test/expected-files.js#L44 and found that it's not `name="file_upload[0]" name="file_upload[1]"` it's just `name="file_upload" name="file_upload"` serverside the file_upload will be an array of 2...

LinusU:
First of all, stellar pull request! Docs, tests and following the coding style, great! :+1: 

Your last example should work with just naming all the files `ordered`. I just realised that it's kind of strange that we are using different rules for fields and for files; maybe we need to do something about that.

I wonder what the implications of using [append-field](https://github.com/LinusU/append-field) on the files would be, I think that there are some safety concerns that we need to address...

How does using the name `file_upload` on all fields suit you?

kylev:
Right, the `array()` version works great if either order doesn't matter, or I can guarantee that the page layout and browser will return the files in some predictable order. However, [browsers don't have any obligation to keep fields in the order they are shown on the page](http://stackoverflow.com/questions/10144206/tell-the-order-of-fields-submitted-from-a-from). So we have to be deterministic about it with clever field names.

kylev:
Yeah the functionality of [append-field](https://github.com/LinusU/node-append-field) is essentially what I'm shooting for...

I could take a shot at getting the above case working, or adding a special form `upload.array('ordered[]')` or similar...

LinusU:
I think that the problem with append-field will be that someone could send something like this:

```text
my_file
my_file[path]
```

And that would probably cause all kinds of problems...

kylev:
I suppose we could leave "your form sucks" behavior unspecified, or reject values that don't follow the convention (i.e., `my_file` wouldn't be a valid file field when array('my_file') is specified. Or, since it is an array, the indexes within the `[]` on field names would have to be both (1) present and (2) integers.

I suspect things will become clearer as I code. But I'm also keeping an eye out for complexity since this will be the first "mapping" in multer, where the name of the field and the name of the place that field ends up are not identical.

LinusU:
@kylev What do you think about #229?

kylev:
Yes, #229 would satisfy my use case.

LinusU:
Closing since #229 got merged

PR #218, opened 2015-09-03
add support for cases where there we don't know exactly which files will be submitted and how many

ophirTot:
adde support to cases where we don’t know which files types will be sent and how many.

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,12 +35,12 @@ Multer.prototype._makeMiddleware = function (fields, fileStrategy) {
     })
 
     function wrappedFileFilter (req, file, cb) {
-      if (fields[0].name != undefined ){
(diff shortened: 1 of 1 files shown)


LinusU:
Indentation is off and there is no documentation. Also, I don't think that this is the right way to solve it. I would prefer #201 and/or a `.any()` function.

ophirTot:
Thanks Linus.

Btw, great job with this project, it works great and very useful.
Can you explain again how can I achieve this with any () function.
I prefer avoiding to fork this project, but we do need to support all types of files and we don't know which one will be sent in advance. 

Thanks for the quick reply.

ophirTot:
It seems to pass all test now

ophirTot:
Do I need to write code for.Any function or is it currently supported?

PR #229, opened 2015-09-19
lib: option to allow any file

LinusU:
Allows the receiving of files regardless of their fieldname.

I'm still not sure if we should merge this. @jpfluger, @hacksparrow, what are your thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,16 @@ Example:
 ]
 ```
 
+#### `.any()`
+
+Accepts all files that comes over the wire. An array of files will be stored in
+`req.files`.
+
+**WARNING:** Make sure that you always handle the files that a user uploads.
+Never add multer as a global middleware since a malicious user could upload
+files to a route that you didn't anticipate. Only use this function on routes
+where you are handling the uploaded files.
+
 ### `storage`
 
 #### `DiskStorage`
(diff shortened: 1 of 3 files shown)


imZack:
Needs this feature :+1:

mjhasbach:
+1

no-donut-for-you:
+1

JedWatson:
I think this is a valuable addition; there are cases in dynamic systems (e.g. CMS) where the field configuration can't be known before a form is submitted, especially when developing a generic API endpoint that may handle updating different types of objects, as per the GraphQL use case.

It seems like the explicit field naming was introduced as a best practice / security feature (?) but for some cases an explicit opt-in to dynamically detect & accept any file fields would be useful.

JedWatson:
@LinusU @jpfluger @hacksparrow any chance we can get an indication of whether this will be merged / implemented? otherwise I'll start working on an alternate solution.

LinusU:
I could probably merge it for you tonight (Swedish time) :cake:

jpfluger:
@LinusU, I'm good with this addition. I also appreciate the **warning** in the readme.

hacksparrow:
👍

JedWatson:
Thanks :100:

LinusU:
Released as `1.1.0` :shipit: :100:

PR #230, opened 2015-09-22
return uploaded files upon done

cdibened:
My thinking is that maybe the developer might need to do some post-processing with the uploadedfiles data.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -48,7 +48,7 @@ function makeMiddleware (setup) {
       drainStream(req)
       busboy.removeAllListeners()
 
-      onFinished(req, function () { next(err) })
+      onFinished(req, function () { next(err, uploadedFiles) })
     }
 
     function indicateDone () {


LinusU:
The files will be add either `req.file` (`single`) or `req.files` (`array`, `fields`, `any`). Why would you need them here as well? Furthermore, this function is meant to use a express middleware so we can't really pass data here.

cdibened:
@LinusU Ah! Dumb thinking on my part.  I completely disregarded the fact you place it in req.file.  I will definitely close this PR.  Thanks for the prompt response.

cdibened:
@LinusU Just a FYI:

Regarding it being middleware, passport.js allows this type of action as well through custom callback.  They pass back two extra pieces of data.
*below text is copied from their site

Custom Callback

If the built-in options are not sufficient for handling an authentication request, a custom callback can be provided to allow the application to handle success or failure.
```javascript
app.get('/login', function(req, res, next) {
  passport.authenticate('local', function(err, user, info) {
    if (err) { return next(err); }
    if (!user) { return res.redirect('/login'); }
    req.logIn(user, function(err) {
      if (err) { return next(err); }
      return res.redirect('/users/' + user.username);
    });
  })(req, res, next);
});
```
In this example, note that authenticate() is called from within the route handler, rather than being used as route middleware. This gives the callback access to the req and res objects through closure.

LinusU:
Hmm, interesting..

PR #249, opened 2015-10-15
Added HTML example to make it more clear

davidgatti:
The server side example alone was unclear that the name of the for needs to match the 'avatar' name. With this additional HTML everything should be clear at a glance. Hope it helps :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,11 @@ Multer adds a `body` object and a `file` or `files` object to the `request` obje
 
 Basic usage example:
 
-### HTML FORM
+## Examples
 
-Singe file:
+### Single file
+
+HTML
 ```html
 <form action="YOUR_URL" method="post" enctype="multipart/form-data">
   <input type="file" name="photo">
@@ -27,16 +29,7 @@ Singe file:
 </form>
 ```
 
-Multiple files:
-```html
-<form action="YOUR_URL" method="post" enctype="multipart/form-data">
-  <input type="file" name="photos" multiple>
-  <input type="submit">
-</form>
-```
-
-### Server side
-
+Server side JavaScript
 ```javascript
 var express = require('express')
 var multer  = require('multer')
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,6 +17,26 @@ Multer adds a `body` object and a `file` or `files` object to the `request` obje
...
+<form action="YOUR_URL" method="post" enctype="multipart/form-data">
+  <input type="file" name="avatar">
+  <input type="submit">
+</form>
+```
+
+Multiple files:
+```html
+<form action="YOUR_URL" method="post" enctype="multipart/form-data">
+  <input type="file" name="avatar" multiple>
```
LinusU: Can we change this name to `photos` instead of changing below? I think it's clearer if they use different names, and it also helps that "photos" is plural and "avatar" is singular

davidgatti:
@LinusU what do you think about now?

LinusU:
What do you think about keeping the first one as `avatar` and having the second one named `photos`? I think it helps users to distinguish them.

davidgatti:
@LinusU There is more text but this to me is super clear and at a glance any new developer will understand how your module work, and best of all he or she can just copy and paste the code :)

What do you think?

micky2be:
Could you add an example purely in Javascript

LinusU:
@davidgatti So sorry for the delay on this, could you remove line 18 as well? And then squash and rebase on master.

@micky2be Do you mean for uploading files from the client? Maybe, although I think that it might be a bit outside of the scope. Could you provide some example code?

micky2be:
For me a JavaScript is more interesting than html one. 
I'll send a PR

davidgatti:
@micky2be but all the rest of the code is in JS. This is a plugin for ExpressJS after all.

davidgatti:
@LinusU why do you want me to squash and rebase? Cant you just merge this pull request?

LinusU:
I prefer to keep the git history clean, as do most other open source projects I worked on. It makes it easier to look at the history.

davidgatti:
This is old, but thankfully GitHub now support squashing, so this is just a reminder of this pull request :)

PR #281, opened 2015-12-03
The WARNING should be at the beginning of the README file!!!

dossan:
I spent a lot of time trying to understand how and when to use multer, and that WARNING helped, but  only when I found it ;) Added an explanation of what 'avatar' is, which made me lose a lot of time...


LinusU:
The warning only applies to the `.any()` function, which is why it was just below it.

Could #249 be a better solution for you?

LinusU:
Closing since I haven't got a response yet, feel free to come with more comments

PR #303, opened 2016-01-20
Added encryptStorage module

MadsRC:
Added support for encryptStorage module and added it to the README.

It uses crypto.createCipher() using the 'AES-256-XTS cipher, with a 36 byte binary key, derived from cypto.randomBytes(). The key is returned to the client in the req.files.encryptKey as a hex string.

With the memory_only boolean (defaults to false) you'll decide to temporarily write to disk before encryption or not.

I'll do some more work on it hopefully verifying that randomBytes actually is a CPRNG, and maybe change it to createCipheriv().

As it is now, it works, and is reasonably secure.

Haven't updated the CHANGELOG or version number.

### DIFF
--- a/storage/encrypt.js
+++ b/storage/encrypt.js
@@ -37,14 +37,13 @@ EncryptStorage.prototype._handleFile = function _handleFile (req, file, cb) {
     that.getFilename(req, file, function (err, filename) {
       if (err) return cb(err)
 
-
       var finalPath = path.join(destination, filename)
       var key = crypto.randomBytes(32).toString('hex')
-      var cipher = crypto.createCipher('aes-256-xts', key);
+      var cipher = crypto.createCipher('aes-256-xts', key)
       if (that.memory_only) {
         var outMemStream = fs.createWriteStream(finalPath)
         file.stream.pipe(cipher).pipe(outMemStream)
-        outMemStream.on('finish', function() {
+        outMemStream.on('finish', function () {
           cb(null, {
             destination: destination,
             filename: filename,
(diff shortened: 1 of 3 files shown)


storage/encrypt.js
```diff
@@ -0,0 +1,93 @@
...
+function EncryptStorage (opts) {
+  this.getFilename = (opts.filename || getFilename)
+
+  if (typeof opts.destination === 'string') {
+    mkdirp.sync(opts.destination)
+    this.getDestination = function ($0, $1, cb) { cb(null, opts.destination) }
+  } else {
+    this.getDestination = (opts.destination || getDestination)
+  }
+  this.memory_only = (opts.memory_only || false)
```
LinusU: Please use camelCase here

storage/encrypt.js
```diff
@@ -0,0 +1,93 @@
...
+  var that = this
+
+  that.getDestination(req, file, function (err, destination) {
+    if (err) return cb(err)
+
+    that.getFilename(req, file, function (err, filename) {
+      if (err) return cb(err)
+
+      var finalPath = path.join(destination, filename)
+      var key = crypto.randomBytes(32).toString('hex')
```
LinusU: This is a potentially blocking call and should probably use a callback.

storage/encrypt.js
```diff
@@ -0,0 +1,93 @@
...
+        })
+      } else {
+        var outStream = fs.createWriteStream(finalPath + '.tmp.original')
+        file.stream.pipe(outStream)
+        outStream.on('error', cb)
+        outStream.on('finish', function () {
+          var input = fs.createReadStream(finalPath + '.tmp.original')
+          var output = fs.createWriteStream(finalPath)
+
+          input.pipe(cipher).pipe(output)
```
LinusU: No error handling on `input` and `output`.

LinusU:
Correct me if I'm wrong, but the `memory_only` option doesn't at all seem to be doing what it says in the documentation?

Also, I'm not sure that this storage engine should live in multer rather than it's own repo, but let's start by reviewing the code and making sure all that is stellar before we make a decision on that...

MadsRC:
I'll take a look at this tonight. Still learning about NodeJS and might have misunderstood how streams work. When I wrote this, I was under the impression that streams required that the entire object fit in memory. Recently, I'm not sure that my original impression is correct. If streams doesn't require the object to fit in memory, the memory_only function would be redundant.

I'll give it a shot tonight and fix it up.

UlisesGascon:
Thanks @MadsRC, and sorry this sat for so long. multer only ships the disk and memory engines; anything else, encryption included, is meant to live as a separate package built on the storage engine contract (see StorageEngine.md), where it can be maintained and reviewed on its own.

I have to say that I love the idea anyway :heart:

PR #318, opened 2016-02-20
maintain object prototype chain for req.body

danielkrainas:
This changes `req.body` to include Object in the prototype.

The `req.body` object provided by expressjs has `Object` in the prototype chain but multer creates an object with no prototype. For code that expects `hasOwnProperty` or another `Object` member, I need to manually set the prototype myself or create a shallow copy to something with `Object` in its chain.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -24,7 +24,7 @@ function makeMiddleware (setup) {
     var fileFilter = options.fileFilter
     var fileStrategy = options.fileStrategy
 
-    req.body = Object.create(null)
+    req.body = Object.create(Object.prototype)
 
     var busboy
 


LinusU:
First of all, great commit message!

This was however done with purpose so I think that we need some discussion before merging this. The problem is that the client data should never be trusted, and thus we can't rely on properties existing up the prototype chain.

Consider what would happen if the client sent `hasOwnProperty=nope` or `hasOwnProperty[call]=`. This would break any script trying to use `req.body.hasOwnProperty(_)`, and having client data break server code is never good.

To be honest, I would be most happy if we could use a `Map` to store the data, but I don't think that people would like there more complicated syntax (`req.body.get('test')` instead of `req.body.test`).

This exact change have actually been proposed before, and I think that my answer then is still relevant: #187

danielkrainas:
While I can respect the cautiousness taken in the design, I believe this is one time where it's better to be consistent as a community, regardless of whether that is right or wrong. Currently, all of bodyparser's parsers spit out an `Object` object:

- [urlencoded](https://github.com/expressjs/body-parser/blob/master/lib/types/urlencoded.js#L84)
- [json](https://github.com/expressjs/body-parser/blob/master/lib/types/json.js#L96)
- [raw](https://github.com/expressjs/body-parser/blob/master/lib/types/raw.js#L60)
- [text](https://github.com/expressjs/body-parser/blob/master/lib/types/text.js#L62)

This creates an expectation at the handler level, regardless of right or wrong, that then gets thrown out when multer is added to the pipeline. What's more is that this is not made clear and a reasonable person wouldn't expect a module like multer to assert an opinion like that, unrelated to its function and purpose.

LinusU:
We feel that keeping it with a `null` prototype is the way to go forward, since the alternative either introduces potential bugs that can be exploited by a client, or have to strip some keys which can be very confusing and hard to work around if it affects you.

If you have anything new to add to the discussion, please feel free to post a comment in the discussions issue: expressjs/discussions#25

Also, as someone pointed out in that issue, `urlencoded` returns a prototype-less object in Node.js 6.x.

Thank you for your time

Issue #320, opened 2016-02-27
expose busboy's options (highWaterMark/fileHwm/defCharset/preservePath)


timbowhite:
Hi. Just spent this morning trying to debug why a particular set of images/fields was causing multer/busboy to hang indefinitely.  Seems that Busboy's Dicer module was not calling the 'end' event after reading a particular field part stream, but only when hanlding an upload of >= 10 total parts, which included 2 particular images.  Very weird, must be an edge case bug in that module.

Anways, I discovered that increasing busboy's `fileHwm` option (in my case >= `53000`) solved the issue.

(This option maps to Stream.Readable's `highWaterMark`: The maximum number of bytes to store in the internal buffer before ceasing to read from the underlying resource.)

So, please expose all of busboy's applicable options (headers can probably be omitted because they're in the req).  I can submit a pull request if preferable.

```
node -v
v0.10.41
```

petersalomonsen:
I would also like to be able to set the preservePath of busboy option via multer. When will this be integrated in the main repository / npm?

damianijr:
@timbowhite you didnt PR about it? I'm on face a problem that this change would be very important.

PR #321, opened 2016-02-27
expose busboy options for #320

timbowhite:
I just followed the module's code style and explicitly specified all the busboy options (but my personal preference would be to just give busboy the options object the user passes with any multer generated options merged in, that way if busboy adds new options, no multer code needs to be changed).


UlisesGascon:
Thanks @timbowhite! This is superseded by #1465

PR #334, opened 2016-04-02
Change line 47 for correctness

milseg:
MyCustomStorage constructor had an attribution to opts object, which was absolutely wrong, since we would like to access getDestination as a method of MyCustomStorage in line 51. Hence the right statement is assigning the value to 'this' which represents the object itself.


LinusU:
Good catch, thanks!

LinusU:
No idea why I never merged this, sorry about that 🎅

PR #372, opened 2016-08-04
Closes #278

### LINKED ISSUE #278: Add .none() to allow multipart but explicitly deny file uploading
I'm using FormData objects as part of my AJAX POST and PUT operations. `xhr.send(new FormData(myForm))` will automatically set the `Content-Type: multipart/form-data`. I'm only sending plain text fields but I still need to parse the multipart body. I've managed to get it working by doing:

```javascript
var upload = multer({dest: __dirname + '/uploads'});
app.put('/whatever', upload.single('SOME_FAKE_FIELD_NAME'), function(req, res) {
    // ...
}
```
I could also use the `.fields()` to set a max count of 0 to fully ensure no files are uploaded, but it would be more convenient is if file uploading could be outright ignored with the following

```javascript
app.put('/whatever', upload.none(), function(req, res) {
```

pavel:
Implementation of #278.

Usage example:
```javascript

var app = express()

app.post("/just-data", upload.none(), function (req, res) {
  // req.body will hold the text fields, if there were any
  // err expected in case of a file upload (https://github.com/expressjs/multer#error-handling)
})
```

### DIFF
--- a/test/none.js
+++ b/test/none.js
@@ -1,0 +1,32 @@
+/* eslint-env mocha */
+
+var assert = require('assert')
+
+var util = require('./_util')
+var multer = require('../')
+var FormData = require('form-data')
+
+describe('None', function () {
+  var parser
+
+  before(function () {
+    parser = multer().none()
+  })
+
+  it('should not allow file uploads', function (done) {
+    var form = new FormData()
+
+    form.append('key1', 'val1')
+    form.append('key2', 'val2')
+    form.append('file', util.file('small0.dat'))
+
+    util.submitForm(parser, form, function (err, req) {
+      // "Unexpected field" error is expected
+      assert.ok(err)
+      assert.deepEqual(req.files, {})
+      assert.equal(req.body['key1'], 'val1')
+      assert.equal(req.body['key2'], 'val2')
+      done()
+    })
+  })
+})
--- a/test/none.js
+++ b/test/none.js
@@ -23,7 +23,7 @@ describe('None', function () {
     util.submitForm(parser, form, function (err, req) {
       // "Unexpected field" error is expected
       assert.equal(err.code, 'LIMIT_UNEXPECTED_FILE')
-      assert.deepEqual(req.files, {})
+      assert.equal(req.files, undefined)
       assert.equal(req.body['key1'], 'val1')
       assert.equal(req.body['key2'], 'val2')
       done()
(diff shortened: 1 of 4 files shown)


test/none.js
```diff
@@ -0,0 +1,32 @@
...
+  it('should not allow file uploads', function (done) {
+    var form = new FormData()
+
+    form.append('key1', 'val1')
+    form.append('key2', 'val2')
+    form.append('file', util.file('small0.dat'))
+
+    util.submitForm(parser, form, function (err, req) {
+      // "Unexpected field" error is expected
+      assert.ok(err)
```
LinusU: Test that the error has the correct `code`

test/none.js
```diff
@@ -0,0 +1,32 @@
...
+      assert.deepEqual(req.files, {})
+      assert.equal(req.body['key1'], 'val1')
+      assert.equal(req.body['key2'], 'val2')
```
LinusU: Split this into two test cases, one without any files, and one with

LinusU:
Could you add a new strategy `NONE` to use instead of `OBJECT` so that `req.files` doesn't get set at all? thx

pavel:
@LinusU Do you want `req.files=null` or `req.files=[]` in case of **"NONE"** strategy?

LinusU:
I don't want anything set at all, in the same way that `VALUE` set's req.file and doesn't touch req.files...

LinusU:
```js
  switch (strategy) {
    case 'NONE': break
    case 'VALUE': break
    case 'ARRAY': req.files = []; break
    case 'OBJECT': req.files = Object.create(null); break
    default: throw new Error('Unknown file strategy: ' + strategy)
  }
```

pavel:
Will it prevent file creation in a `dest` designated folder?

test/none.js
```diff
@@ -0,0 +1,47 @@
...
+
+  it('should not allow file uploads', function (done) {
+    var form = new FormData()
+
+    form.append('key1', 'val1')
+    form.append('key2', 'val2')
+    form.append('file', util.file('small0.dat'))
+
+    util.submitForm(parser, form, function (err, req) {
+      // "Unexpected field" error is expected
```
LinusU: Can you remove this comment and add `assert.ok(err)` as well so that it gives a better assertion error if no error was given. (as it is now it will give "Cannot read property code of undefined")

LinusU:
Yes, no files should be created. One last comment, then we are good to go. Thank you for your contribution!

LinusU:
Thanks!

pavel:
You're welcome.

LinusU:
Released as 1.2.0

jpfluger:
Good job guys!  :+1:

PR #376, opened 2016-08-25
Check and discard array notation from fieldname when fileStrategy is ARRAY

JasonMore:
I noticed angular.js was adding an array notation to the fieldname, and this check removes that. Not sure if this is valid or not but thought I would share since I'm sure others have come across this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,6 +35,11 @@ Multer.prototype._makeMiddleware = function (fields, fileStrategy) {
     })
 
     function wrappedFileFilter (req, file, cb) {
+      if (fileStrategy === 'ARRAY'){
+        // remove array notation: ex "file[0]" to "file"
+        file.fieldname = file.fieldname.replace(/\[\d\]/gi, '')
+      }
+
       if ((filesLeft[file.fieldname] || 0) <= 0) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -66,10 +71,6 @@ Multer.prototype.fields = function (fields) {
   return this._makeMiddleware(fields, 'OBJECT')
 }
 
-Multer.prototype.none = function () {
```
LinusU: Please don't remove `.none`

LinusU:
I'm currently on vacation, will do proper review when I get home...

JasonMore:
Oh sorry that was an accident. I'll fix it and force push the actual 3 line change

JasonMore:
@LinusU fixed that bad rebase. Correct 3 lines of code show up in the diff now

starcwl:
I find this problem when I use ng-file-upload.
so I change code to

`      var fieldname = (fileStrategy === 'ARRAY') ? file.fieldname.replace(/\[\d+\]$/,"") : file.fieldname;

      if ((filesLeft[fieldname] || 0) <= 0) {
        return cb(makeError('LIMIT_UNEXPECTED_FILE', file.fieldname))
      }
`

Issue #335, opened 2016-04-04
Crash if multiple files have errors. 


Gregroam:
For example, if you try to upload 2 files that are both too big, you get a weird uncaught  error
`TypeError: path must be a string
    at TypeError (native)
    at Object.fs.unlink (fs.js:927:11)`

And this seems to be because on the first error, it unlinks the uploaded files, and deletes their filepath and other properties, then on the 2nd error, the filepath property does not exist on the uploaded files, and so this weird uncaught error occurs and crashes the process.

RafLeszczynski:
have the same issue

Roam-Cooper:
Can @LinusU ping us back as to whether they know this issue has been resolved or not, otherwise we can fix via a pull request as we are using this module for a project that will go to production in a few months.

LinusU:
I haven't been able to reproduce this myself so I don't know if it's present or not. If someone could submit a test case that fails I would be more than happy to fix it :)

Gregroam:
Node app
```
var express = require("express");
var app = express();

var multer = require("multer");
var upload = multer({
	dest: __dirname+"/upload",
	limits: {fileSize: 10000}
});

app.post("/upload", function(req, res, next) {
	return upload.array("file")(req,res, function(err){
		if (err) {
			console.log("error",err);
			return res.end("Failed to upload");
		}
		console.log(req.files);
		return res.end("Uploaded sucessfully");
	});
})

app.listen(8080, function() {
	console.log("Listening");
})
```

[Postman Screenshot of request body with 1 file](http://i.imgur.com/HATMi0z.png)
When I try to upload only one file greater than 10,000 bytes. Then I get "Failed to upload" and the console logs as expected:
 `error { [Error: File too large] code: 'LIMIT_FILE_SIZE', field: 'file', storageErrors: [] }`


[Postman Screenshot of request body with 2 files](http://i.imgur.com/D5Npu1t.png)
When I try to upload two files greater than 10,000 bytes each. Then the process crashes with this error.
```
fs.js:927
  binding.unlink(pathModule._makeLong(path), req);
          ^

TypeError: path must be a string
    at TypeError (native)
    at Object.fs.unlink (fs.js:927:11)
    at DiskStorage._removeFile (/Test/node_modules/multer/storage/disk.js:61:6)
    at remove (/Test/node_modules/multer/lib/make-middleware.js:64:19)
    at handleFile (/Test/node_modules/multer/lib/remove-uploaded-files.js:10:5)
[...]

PR #380, opened 2016-08-31
Prevent Multiple Errors from Crashing

AdrianSchneider:
When you have an upload that has more than one error, seemingly only with disk storage, it causes the node process to crash. More details here - https://github.com/expressjs/multer/issues/335

```
Uncaught TypeError: path must be a string
```

Tests should showcase an example of this, along with a potential fix.


Enether:
Why is this not in the latest version?!

LinusU:
Sorry, must have missed this, merging now

PR #388, opened 2016-09-16
Update package.json

wtonsp:
Update busboy to 0.2.13.  Updated in my environmen to support nexe properly.


ctcpip:
newer busboy used now

PR #356, opened 2016-06-24
#211 Add a transformFile option

wesleytodd:
This is obviously not a PR ready for merging.  But I wanted to post it to get the discussion started.  Before I write tests and docs, is there anything inherently wrong with the approach I took on this?  Is there any problem with not handling the `fileStream` events?  Or would that be up to the implementor of the transform method?

This is an example of how I saw this working from a users perspective:

``` javascript
var upload = multer({
  transformFile: function (req, file, cb) {
    file.stream = pump(file.stream, resizeMyImage(file), saveMyImageSomewhereElse(file))
    // Optionally change other parts of the file object...
    cb(null, file)
  }
})
```

Thoughts? Feedback before I write tests to make sure it all works correctly?

Issue #182, opened 2015-07-27
Introduce stricter default limits


LinusU:
Many limits are currently defaulting to unlimited which isn't that smart at all. I think that we should gather some data on what other libraries are doing and then figure out some good, rather restrictive limits.

I don't see a problem with keeping the low since they are very easy to increase for the user, and since the error message is really clear when a file is to big.

ryhinchey:
@LinusU I did a little investigation and here's a breakdown of what I found:

https://github.com/node-formidable/node-formidable/blob/master/src/Formidable.js#L14-L22
https://github.com/pillarjs/multiparty/blob/master/index.js#L61-L63

Limits
`fields`:  `multiparty`: 100, `node-formidable`: 1000
`fileSize`: `multiparty`: Infinity, `node-formidable`: 200mb

If you're still interested in adding stricter default limits, I'd be happy to make a PR

LinusU:
Thanks! I wish I'd thought of this issue a bit sooner, since this is a breaking change...

I'll try to add this to #399 asap!

ryhinchey:
@LinusU I made a PR for the change https://github.com/expressjs/multer/pull/848

PR #399, opened 2016-09-29
Switch to stream based API

LinusU:
As have been discussed in a few other issues, I thought it would be interested to see what the api would look like if Multer just gave you `.stream` on the uploaded files. This removes the concept of storage engines completely and you can use any other module that works with streams, instead of multer-specific modules.

I would love some thoughts and feedback on this ❤️ 

ping @expressjs/multer-collaborators, @wesleytodd


wesleytodd:
Love this idea.  It is much simpler, more modular, and would solve way more use cases than the current situation (#356 for example).  I don't have time now to read all the code, but in theory this has my thumbs up.  I will try to look through it more throughly this weekend.

niftylettuce:
My only issue @LinusU is to fix it so that you expose `_makeMiddleware` again because it's needed in `koa-multer` here <https://github.com/koa-modules/multer/blob/master/index.js#L19>.  Here was the commit it was removed in <https://github.com/expressjs/multer/commit/42a9e3b6376ddd9878f2eb26060713e7479ccb3f>.

LinusU:
Just released `2.0.0-alpha.1` for you guys to try it out. It's published under the `next`-tag and thus won't be automatically installed for anyone.

🎉

LinusU:
@niftylettuce I'll look into the koa-issue shortly 👌

LinusU:
@niftylettuce see koa-modules/multer#11

niftylettuce:
You are a rockstar.  How do I donate to you?

niftylettuce:
Also @LinusU - is limits option still enforced?

LinusU:
> How do I donate to you?

There are others who need it more than I, e.g. https://my.charitywater.org/donate/home

> is limits option still enforced?

Yes they are :ok_hand:

LinusU:
`2.0.0-alpha.2` is out with improvements to the error codes. Happy coding :tada: 

To try it out, use the following command: `npm install --save multer@next`

hacksparrow:
👍🎉👍

niftylettuce:
It's working great for me!  I have a fork of `koa-multer` with this version added and @LinusU fork fix https://github.com/niftylettuce/koa-multer/tree/lu-fix

qwemaze:
Solves many problems, thanks!

But I don't understand why no encoding and mimetype properties in a file? I think it's critical.

qwemaze:
Oh nevermind, mime type is usually determined by file's extension, and file encoding would not help at anything. Correct if wrong.

LinusU:
That is correct, as I have understood it the encoding is only how it's sent over the wire and doesn't matter at all for the end user

lib/middleware.js
```diff
@@ -0,0 +1,43 @@
...
+          appendField(req.body, field.key, field.value)
+        })
+
+        var appendFile = createFileAppender(options.fileStrategy, req, options.fields)
+
+        result.files.forEach(function (file) {
+          var stream = fs.createReadStream(file.path)
+
+          stream.on('end', function () {
+            fs.unlink(file.path, function () {})
```
LinusU: Actually, we can unlink this file as soon as it has been opened, that way the file wills still be removed even if it isn't consumed. More info: http://stackoverflow.com/a/3181726/148072

LinusU:
@dougwilson Would you mind giving this a quick look? 💌 

I would like to start moving towards releasing this :)

hacksparrow:
@LinusU `CHANGELOG` needs an update.

LinusU:
- [x] changelog updated

LinusU:
Since it will help people keeping their apps secure, and since a lot of the opened issues are about it, I'm leaning towards including mime-detection with 2.x. This is my current thoughts:

The `File` object would gain the following fields:

```js
// The detected mime-type, or null if we failed to detect
detectedMimeType: String | Null

// The typical file extension for files of the detected type, or null if we failed to
// detect (with leading `.` to match `path.extname`)
detectedFileExtension: String | Null

// The mime type reported by the client using the `Content-Type` header, or null if the
// header was absent
clientReportedMimeType: String | Null

// The extension of the file uploaded (as reported by `path.extname`)
clientReportedFileExtension: String
```

Any thoughts on this would be appreciated; anyone is welcome to give feedback! :)

wesleytodd:
@LinusU I think this is a great idea, as I imaging almost everyone needs most of this info anyway.  I know I do in my apps.

LinusU:
Note to self: When releasing this, remember to update this page: http://expressjs.com/en/resources/middleware/multer.html

LinusU:
Released `multer@2.0.0-alpha.4` with file type detection support 🎉

LinusU:
Released `multer@2.0.0-alpha.5` with a fix for #455 🎉

LinusU:
Released `multer@2.0.0-alpha.6` with a new error code that will be given when a client aborts a request. Related to #438

oshalygin:
Fairly stupid question because I'm not terribly familiar with multer other than whats on the docs, but what is the API to consume the stream?  I essentially want to proxy the multipart file upload request to another backend and would like to take the stream and pass it along

wesleytodd:
@oshalygin You can see the docs here: https://nodejs.org/api/stream.html

But most likely you just want to `pipe` the entire request if you want to maintain the multipart form encoding.  In which case you wouldn't need this module.  If on the other hand, you want to pipe just one file to the upstream, then you could decode the multipart form with this module, and just write the one file to the upstream.

oshalygin:
@wesleytodd awesome thanks I'm going to give that a whirl.

orenklein:
I like this idea of switching multer API to stream. :)

Few thoughts regarding multer new api implementation:
Are you creating a temporary file for each file field and then creating read stream from that file? (https://github.com/expressjs/multer/blob/explore-new-api/lib/read-body.js#L71)
If so, couldn't we get the same outcome with current multer using DiskStorage - createReadStream and then remove/unlink created file?
(https://github.com/expressjs/multer/blob/explore-new-api/lib/middleware.js#L25)
Furthermore, what if someone would prefer to stream/pipe uploaded file through memory instead of a temporary file?

Last one: what are your thoughts about an option to pipe the request's stream file directly?
(Without writing anything to a temporary file or memory, although I guess that's not the purpose of this library).
@LinusU

LinusU:
Hi @orenklein, sorry for the late reply :(

> Are you creating a temporary file for each file field and then creating read stream from that file? If so, couldn't we get the same outcome with current multer using DiskStorage - createReadStream and then remove/unlink created file?

That is probably doable, although here it would be done automatically for the user...

> Furthermore, what if someone would prefer to stream/pipe uploaded file through memory instead of a temporary file?

My plan is to let this decide itself, so that small files only ends up in memory, and large files on disk. Since the new api completely hides this implementation detail, it could be added at a later date without anything needing to be changed for the user!

> Last one: what are your thoughts about an option to pipe the request's stream file directly?
(Without writing anything to a temporary file or memory, although I guess that's not the purpose of this library).

I think that this use case is better served by busboy, or by having a lower level api that we in itself build Multer on top of

niftylettuce:
@LinusU Can you publish a new package with this?  It doesn't seem like multer will get your new strema based API 😢

niftylettuce:
Actually I mean - why don't we do this directly in a new major version of multer?

LinusU:
It's currently published as `multer@2.0.0-alpha.6`, the plan is to make it `2.0.0` as soon as I feel comfortable with the final api :)

glortho:
Great to see this! Any forecast when multer 2.0.0 will come out of alpha?

jmonfort:
@LinusU Hey!
I was hoping to use multer for a file upload proxy.
Any update on when 2.0 will be released?

Thanks!

LinusU:
Hey 👋 

Really hoping to get to this soon, but unfortunately, I only have so much time to spend on open source.

Trying it out and reporting any issues (or even better, reporting that it works bug free ☺️) here would absolutely help to get me more comfortable to release it out of alpha. I don't currently feel that there is anything missing that needs to be added, just need to feel comfortable to push it out to all users ☺️

jmonfort:
I've just tested and it works fine. Anxious to use it.
The few changes that caught me by surprised were:
- File.originalname => File.originalName
- File.mimetype => File.detectedMimeType
- File.buffer is no longer available. 

Otherwise, it's working fine.
Thanks for the reply

LinusU:
Published 2.0.0-alpha.7 that drops support for Node.js <6, and bumps all dependencies! 🚀

niftylettuce:
Wooooot!

niftylettuce:
Stream based API ready? 😄

niftylettuce:
Can we get this on master branch?

niftylettuce:
And move it out of alpha?

index.js
```diff
@@ -1,100 +1,60 @@
...
-  return makeMiddleware(setup.bind(this))
+  return createMiddleware(setup.bind(this))
 }
 
 function multer (options) {
-  if (options === undefined) {
-    return new Multer({})
-  }
+  if (options === undefined) options = {}
+  if (options === null) throw new TypeError('Expected object for arugment "options", got null')
```
jdleesmiller:
```suggestion
  if (options === null) throw new TypeError('Expected object for argument "options", got null')
```

index.js
```diff
@@ -1,100 +1,60 @@
...
+  if (typeof options !== 'object') throw new TypeError('Expected object for arugment "options", got ' + (typeof options))
```
jdleesmiller:
```suggestion
  if (typeof options !== 'object') throw new TypeError('Expected object for argument "options", got ' + (typeof options))
```

LinusU:
@jdleesmiller thanks!

I'll try to get some time this week to:

1. fix the typos
2. drop support for Node.js 6 (it's end of life)
3. bump dependencies
4. cut an RC release

🚀

bmoe24x:
Is there any update on timeline to alpha being merged to master? My workflow requirements are not met (at least without some 'hacking') by master version, only alpha, as I need to pre-process files before storing, see if file limit is exceeded, etc.

Thanks!

exentrich:
No update?

LinusU:
I've just released `v2.0.0-beta.1` :shipit: 🎉 

It drops support for Node.js <8.3, so be sure to upgrade to at least Node.js 8.3.0 (preferably 12.13.1) before upgrading.

I've also started tested on 1) the minimum supported version of Node.js (`8.3.0`) in addition to latest major, and 2) on both the macOS and Windows platforms in addition to Linux. Hopefully this helps keeping all builds really stable! 🙌 

As usual, you can install it like so:

```sh
npm install --save multer@next
```

-----

I would love to get some reports from people using this in a production setting! It would really help me to know that this has been thoroughly tested and is ready to be tagged as `2.0.0`.

(ping @bmoe24x @qwelias)

niftylettuce:
@LinusU 	can you please release 2.0.0 publicly out of alpha?  thank you

LinusU:
Hello everyone 👋 

Thank you for your patient while I slowly work on this, I'm sorry that I haven't been moving faster forward with this.

I just released the release candidate for version 2.0.0 (`2.0.0-rc.1`).

Unless anyone discovers something worth blocking the release for, I will release this exact version as 2.0.0 on the first of March!  🚀 **2020-03-01** 🚀 

I've added this to my calendar so there should be no further delays.

A huge thank you to everyone that's been involved with this! ❤️ 

-----

Changelog for 2.x:

- **Breaking:** switch to stream based API
- **Breaking:** drop support for Node.js < 10.13.x
- **Feature:** unlink file as soon as it's opened
- **Feature:** add file type detection
- **Fix:** allow files without filename
- **Fix:** handle client aborting request
- **Internal:** achieve 100% code coverage 💪 
- **Internal:** test on macOS & Windows ✨

bwgjoseph:
Hi, would `@types/multer` for v2 be release together?

LinusU:
Sorry for not getting this out as promised 😬 

I just want to merge in #848, cut a new release candidate, and wait a few days before promoting that.

Will try and merge that one this afternoon as soon as I'm finished with work!

LinusU:
@bwgjoseph the types are managed in a separate repo, but if I have time for it I will submit a PR there as well 👍 

Realistically it will probably take a few days before it's out, efter Multer 2.0 has been released...

LinusU:
Okay, I've changed some things concerning the limits in an attempt to address #182 

- Allow limits to be passed as string, e.g. `'12MB'`
- Remove `parts` limit in favour of `fields` & `files`
- Set default values for all limits
- Lower limits to reasonable defaults

The last part was very very hard, since this is used in a variety of applications it's impossible to tell what reasonable is. I'm actually leaning towards requiring limits to be specified, as that would force each app to decide it's own limits. Any thoughts on this?

These are the new limits:

Key | Description | Default
--- | --- | ---
`fieldNameSize` | Max field name size | `'100B'`
`fieldSize` | Max field value size | `'8KB'`
`fields` | Max number of non-file fields | `1000`
`fileSize` | The max file size | `'8MB'`
`files` | The max number of file fields | `10`
`headerPairs` | The max number of header key=>value pairs to parse | `2000` (same as Node's http)

The maximum body that one can squeeze thru now should be `< 100MB`:

[((8KiB + 100B) * 1000) + (8MiB * 10) = 88MB](https://www.wolframalpha.com/input/?i=%28%288KiB+%2B+100B%29+*+1000%29+%2B+%288MiB+*+10%29)

I'll push out this as Release Candidate 2 now 🚀

bwgjoseph:
Hi,

Was wondering if the description for `fileSize` could be described as `The max size per file`. It could be just me, but I was thinking to myself if it is `10 (file) / 8` which gets me the `filesize` per file or the other way round. 

I guess the same goes for `fieldNameSize` and `fieldSize`

index.js
```diff
@@ -12,29 +14,29 @@ function _middleware (limits, fields, fileStrategy) {
 
 class Multer {
   constructor (options) {
-    this.limits = options.limits
+    this[kLimits] = options.limits
```
ryhinchey: out of curiosity, what's the benefit of using a  `Symbol` here?

ryhinchey:
Also this should now close https://github.com/expressjs/multer/issues/851

index.js
LinusU: This was made so that the limits can't be seen/modified from the outside.

I don't really have anything against that, but if it's possible it should be documented, and since it wasn't possible before I don't think that there is a use case for it.

If we leave it in but don't document, then my fear is that we might break it in a future update without meaning to ☺️

LinusU:
@bwgjoseph great feedback, I'll add that!

index.js
ryhinchey: makes complete sense! thank you

bwgjoseph:
H-i, was wondering how do you detect the mimeType that is return in `detectedMimeType`? I tested a couple of files and it all return as `null`

Like `.md`, `.js`, `.json` and even `.txt` file.

README.md
```diff
@@ -65,45 +66,27 @@ app.post('/profile', upload.array(), function (req, res, next) {
...
-The following are the options that can be passed to Multer.
-
 Key | Description
 --- | ---
-`dest` or `storage` | Where to store the files
-`fileFilter` | Function to control which files are accepted
-`limits` | Limits of the uploaded data
+`fieldName` | Field name specified in the form
+`originalName` | Name of the file on the user's computer (`undefined` if no filename was supplied by the client)
+`size` | Size of the file in bytes
```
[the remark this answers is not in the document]
LinusU: This will be the total size of this file, updated the documentation to clarify that 👍 

re. the wider question, all that will have been received by the time that the Multer middleware handles control back to Express via the `next` function ✔️

LinusU:
@bwgjoseph we are using the [`file-type` package](https://github.com/sindresorhus/file-type)

Unfortunately, detecting the file type of text based formats are very hard, since there isn't much to go on.

e.g. a `.txt` file can contain any text, so we cannot really "detect" it. An `.md` Markdown file is also really hard, and would certainly have many false positives (in fact, I think that all text files are valid Markdown files?)

bwgjoseph:
@LinusU, apologies as I didn't read much beforehand. My example isn't really great, I see that you're using `file-type` so I can just dive into that to see what they provide and what not. But thanks!

LinusU:
Hi everyone, sorry for not having released a final version yet! I just realised that there is one last semi-breaking change that I want out. I'll try and do this tonight and push a new RC.

I want to go thru and improve all the error messages, and give them proper attributes for all relevant data. This seems to be a huge source of confusing for some people, especially the "unexpected field" one, so I want to improve on that!

lib/read-body.js
```diff
@@ -0,0 +1,135 @@
...
+  stream.on('readable', stream.read.bind(stream))
+}
+
+function collectFields (busboy, limits) {
+  return new Promise(function (resolve, reject) {
+    var result = []
+
+    busboy.on('field', function (fieldname, value, fieldnameTruncated, valueTruncated) {
+      if (fieldnameTruncated) return reject(new MulterError('LIMIT_FIELD_KEY'))
+      if (valueTruncated) return reject(new MulterError('LIMIT_FIELD_VALUE', fieldname))
```
noonii, line 25: When an error is thrown from `field` event, where are the uploaded files removed? Often files are parsed before fields.

bwgjoseph:
Hi @LinusU, any further update to this release?

prateekkathal:
Waiting on this 😞

LinusU:
@bwgjoseph, @prateekkathal I've published a new RC that fixes compatibility with Node.js 16 🚀 

The majority of our dependencies (and the ecosystem at large) is moving to ESM over CommonJS. In order to be able to import those packages (e.g. fs-temp 2.0 which fixes Node.js 16 compatibility) we also need to be an ESM package.

Potentially we could work around this if this is too much of a disruption for people, but I personally feel that this is the way forward for the entire Node.js ecosystem 🚀 

Here is the changelog for 2.0.0-rc.3:

> Migration Guide:
>
> This relases changes the package from a Common JS module to an EcmaScript module, and drops support for older versions of Node.
>
> - The minimum version of Node.js supported is now: `12.20.0`, `14.13.1`, and `16.0.0`
> - The package must now be imported using the native `import` syntax instead of with `require`

Any and all feedback is much welcomed!

jonchurch:
It's pretty far out from stable, but [WHATWG streams](https://nodejs.org/api/webstreams.html#webstreams_web_streams_api) seem like they'd be a nice addition here eventually. That will probably be Multer 3.0, though!

bwgjoseph:
Hi @LinusU,

I was testing on one of my (typescript) project, and it seem to throw some ESM error while running with the latest `multer` (rc3)

```log
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: \feathers-ottoman-demo\node_modules\multer\index.js   
require() of ES modules is not supported.
require() of \feathers-ottoman-demo\node_modules\multer\index.js from \feathers-ottoman-demo\src\services\attachments\attachments.service.ts is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from \feathers-ottoman-demo\node_modules\multer\package.json.

    at Module._extensions..js (internal/modules/cjs/loader.js:1080:13)
    at Object.nodeDevHook [as .js] (\feathers-ottoman-demo\node_modules\ts-node-dev\lib\hook.js:63:13)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (\feathers-ottoman-demo\src\services\attachments\attachments.service.ts:6:1)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Module._compile (\feathers-ottoman-demo\node_modules\source-map-support\source-map-support.js:568:25)      
    at Module.m._compile (C:\Users\Joseph\AppData\Local\Temp\ts-node-dev-hook-005731216882142842.js:69:33)
[ERROR] 14:17:26 Error: Must use import to load ES Module: \feathers-ottoman-demo\node_modules\multer\index.js    
require() of ES modules is not supported.
require() of \feathers-ottoman-demo\node_modules\multer\index.js from \feathers-ottoman-demo\src\services\attachments\attachments.service.ts is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from \feathers-ottoman-demo\node_modules\multer\package.json.
```

This is my `tsconfig` which I added `ts-node` configuration

```json
{
  "ts-node": {
    // See https://github.com/TypeStrong/ts-node/issues/922
    // these options are overrides used only by ts-node
    // same as our --compilerOptions flag and our TS_NODE_COMPILER_OPTIONS environment variable
    "compilerOptions": {
      "module": "commonjs"
    }
  },
  "compilerOptions": {
    "target": "es2018",
    "module": "commonjs",
    "outDir": "./lib",
    "rootDir": "./src",
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true
  },
  "exclude": [
    "test"
  ]
}
```

And I'm using `multer` like such

```ts
import multer from 'multer';

const MAX_FILE_SIZE = 5242880;
const MAX_FILE_UPLOAD_LIMIT = 5;

// Instantiate multer
const multerware = multer({
  limits: {
    fileSize: MAX_FILE_SIZE,
    files: MAX_FILE_UPLOAD_LIMIT,
  },
});

multerware.array('files'),
```

If I add `type: 'module` in `package.json`, it doesn't change anything. The same error will be thrown.

Any idea?

bwgjoseph:
Hey @LinusU,

Thanks for helping me out, appreciate it. 

I still am not able to get it working properly even with the different trials and errors. I think I don't understand enough of all these, will do up more reading first. So I reverted back to `rc2` for now.

 And I also don't want to clutter this thread further more if this is not a `multer` issue

Cheers!

mohe2015:
Maybe use this opportunity to update all packages to their new major versions? And maybe remove mkdirp and replace it with its native counterpart https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#fs_fspromises_mkdir_path_options / https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#fsmkdirpath-options-callback

LinusU:
> Still the conflicts probably need to be resolved at some point but I don't know if this would be needed now.

When we release version 2.0.0 I will rename the `master` branch to `v1.x` and this branch to `master`

niftylettuce:
Bump 😉

bryanph:
@LinusU How do you feel about just releasing the 2.0 and seeing how it goes? That way more people get to try it and any bugs that might exist in it will surface quicker. Also gives the community a chance to contribute if bugs are found. What do you think?

In any case thanks for this project, it's helped me out a lot! 😁

PR #400, opened 2016-09-30
add chinese language readme.md

zengming00:
add chinese language readme.md, to let chinese better use this project , tks

### DIFF
--- a/README-zh-cn.md
+++ b/README-zh-cn.md
@@ -1,0 +1,248 @@
+# Multer [![Build Status](https://travis-ci.org/expressjs/multer.svg?branch=master)](https://travis-ci.org/expressjs/multer) [![NPM version](https://badge.fury.io/js/multer.svg)](https://badge.fury.io/js/multer) [![js-standard-style](https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat)](https://github.com/feross/standard)
+
+Multer 是一个 node.js 中间件，用于处理 `multipart/form-data` 类型的表单数据, 它主要用于上传文件. 它是写在 [busboy](https://github.com/mscdex/busboy) 之上非常高效。
+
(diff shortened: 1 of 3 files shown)


LinusU:
Thank you for this, it is awesome. The only concern I have is how we will keep this up to date 💭 

I would like to bring this up with the Express TC and see what the thoughts are...

jpfluger:
Nice job @zengming00! :+1:  Keeping the translation up-to-date is the issue, though. I do like @crandmck's observations in [Translated Readmes](https://github.com/expressjs/discussions/issues/38). Maybe...? 

* Place translations into a subfolder for translations
* Each translation gets a disclaimer at the top
* Each translation at the top displays the multer version to which the translation applies.

zengming00:
@jpfluger good idea,  yes, keeping the translation up-to-date is the issue, and @crandmck is right , small differences in translation can completely change the meaning, and i'm non-English speakers, In this document, have some sentences real can't understand, I use myself experience and coding some demo to try it , so translate to other language it's not an easy thing. may be my translation can help some people.
How to keep this up to date , may be i can period update it , a week or a month , or other chinese people will be update it . tks

crandmck:
> Each translation gets a disclaimer at the top
> Each translation at the top displays the multer version to which the translation applies.

@jpfluger Yes, I think these suggestions make a lot of sense.  As a convenience, I would also add the translation date (even though someone could figure that out from the git history).

So @zengming00 could you please add the following note **in Chinese** at the top of the page:

> This document was translated on 3 Oct 2016 for multer version 1.2.0. It may be outdated compared to the [English README](README.md). 

I don't know that the Chinese README needs to be in a separate folder...  Since the filename has the ISO lang code, that might be sufficient.  But I'll leave that decision to @LinusU and others.

zengming00:
@crandmck updated

crandmck:
LGTM.

LinusU:
Awesome, thank you for this!

PR #410, opened 2016-10-20
Keep file encoding and mimetype info

qwemaze:


qwemaze:
Oh nevermind, mime type is usually determined by file's extension, and file encoding would not help at anything. Correct if wrong.

PR #429, opened 2016-12-13
Abort file upload on request 'close' event

### LINKED ISSUE #259: file is not removed if upload is canceled
Consider this minimal working example (Multer v1.1.0 on Windows 7, OS X 10.11 and Ubuntu 14.04 with node v4.2.1):

``` js
var app = express();

app.post("/upload", upload.single("file"), function(req, res, next){
    res.end("uploaded to: " + req.file.filename);
});
app.listen(80);
```

and the following test script

``` bash
# create large test file
$ dd if=/dev/urandom bs=2G count=1 of=upload
# upload
$ curl -F "file=@upload" http://localhost/upload
# upload again, but cancel before upload is finished
timeout -s SIGKILL 2s curl -F "file=@upload" http://localhost/upload
```

**Expected behavior:**
Only the first file is persisted, the second is removed after the upload got canceled.

**Actual behaviour:**
Both files are stored in the _uploads_ folder

adskjohn:
fixes #259

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -139,9 +139,9 @@ function makeMiddleware (setup) {
           abortWithCode('LIMIT_FILE_SIZE', fieldname)
         })
 
-        var closed = false;
+        var closed = false
         req.on('close', function () {
-          closed = true;
+          closed = true
         })
 
         storage._handleFile(req, file, function (err, info) {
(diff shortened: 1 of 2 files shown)


zlatinejc:
I do not think looking at disk events will solve this issue.

You should be focusing on incoming network stream.

UlisesGascon:
Thanks @adskjohn! This was fixed in multer 2.3.0 as part of https://github.com/expressjs/multer/security/advisories/GHSA-qfvm-cv95-jqjf, so I'm closing this one.

PR #431, opened 2016-12-14
disk.js update

Piotr23:
creating directories for diskStorage manual invoke too


Piotr23:
creating directories for diskStorage manual invoke too with test case change

PR #432, opened 2016-12-14
creating path folders by default for using multer.diskStorage

Piotr23:


Piotr23:
is there someone?

LinusU:
Sorry for not seeing this earlier.

I don't think we should do sync IO that's not on the first tick. Could you change to using the async version of mkdirp instead?

jonchurch:
Is this still something that should be implemented? 

From what I understand, this will create folders for paths when `destination` is a function. So the new behavior would be that any path returned by `destination` would be created for users if it doesn't exist already.

UlisesGascon:
Thanks @Piotr23, and sorry this waited so long. We're keeping the documented behaviour: multer creates the folder when `destination` is a string, and leaves it to your function when `destination` is a function, since a per-request path is yours to validate and creating whatever it returns would hide mistakes. Inside that function it's one line: `fs.mkdir(dir, { recursive: true }, function (err) { cb(err, dir) })`.

PR #447, opened 2017-01-19
Close request on file size limit reached. Fixes #344

### LINKED ISSUE #344: File size limit cancels file upload only after all bytes are received
I've been using multer for file uploads, and I've noticed an issue with the file size limit.  I'm using the limits option to prevent the upload of files that are too large.

```javascript
  const upload = multer({
    storage: multer.memoryStorage(),
    limits: {
      fieldNameSize: 255,
      fileSize: 500000,
      files: 1,
      fields: 1
    }
  });
```

When uploading an image to my API, multer seems to wait until it has received all of the image bytes before failing with the "LIMIT_FILE_SIZE" code.

Isn't it a security vulnerability if your API allows someone to upload an arbitrarily large file _before_ failing the request?  Is there another technique for preventing the behavior?

DiegoRBaquero:
Busboy only emits the 'limit' event. And even if it unpipes the multipart upload to ReadableStream file, it doesn't closes it.

This change emits the 'close' event so that the DiskStorage closes the WriteableStream to the temporal disk file and immediately calls the callback that allows the temp file to be removed and callbacks to express.

It uses outStream.close() instead of end() to maintain the convention that finish event should be upon full successful write only. Thus the cb is called.

Took about 4 hours to debug. But works.

One important thing to note is that if you are using a reverse proxy, buffering MUST be disabled and error 413 should be the response. Sending 400 wouldn't abort the request in nginx.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -136,6 +136,7 @@ function makeMiddleware (setup) {
         })
 
         fileStream.on('limit', function () {
+          fileStream.emit('close')
           aborting = true
           abortWithCode('LIMIT_FILE_SIZE', fieldname)
         })
--- a/storage/disk.js
+++ b/storage/disk.js
@@ -38,15 +38,21 @@ DiskStorage.prototype._handleFile = function _handleFile (req, file, cb) {
       var outStream = fs.createWriteStream(finalPath)
 
       file.stream.pipe(outStream)
+      file.stream.on('close', function () {
+        outStream.close()
+        onFinish()
+      })
       outStream.on('error', cb)
-      outStream.on('finish', function () {
+      outStream.on('finish', onFinish)
+      
+      function onFinish () {
         cb(null, {
           destination: destination,
           filename: filename,
           path: finalPath,
           size: outStream.bytesWritten
         })
-      })
+      }
     })
   })
 }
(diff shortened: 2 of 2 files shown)


lib/make-middleware.js
```diff
@@ -136,6 +136,7 @@ function makeMiddleware (setup) {
         })
 
         fileStream.on('limit', function () {
+          fileStream.emit('close')
```
LinusU: I think that theses events are only meant to be emitted internally by the streams. The correct way would probably be to `unpipe` the incoming request stream 🤔

goatandsheep: Is there any way for someone else to make this change, since @DiegoRBaquero seems busy?

SuperOP535: I need this

DiegoRBaquero: @LinusU would you mind making the change? I couldn't figure it out. change on disk would be to listen to `unpipe` event instead of close.

LinusU: I'm not sure exactly what should be changed though, I think that this requires a lot of testing to get right 🤔

goatandsheep:
Was the feature implemented? Why was this closed?

PR #451, opened 2017-01-25
Add file type detection

LinusU:
Unfortunately we must drop support for Node.js 0.12 since [`file-type`](https://github.com/sindresorhus/file-type) doesn't work on it 😞 

As discussed on last Express TC meeting though, since we have already dropped 0.10, it shouldn't be a big deal. If we drop 0.10 we might as well drop 0.12 also...


LinusU:
Thanks for the review @cstruct 👍

PR #476, opened 2017-04-14
Add transform option and throttle test

kapouer:
This adds the ability to transform stream before handing it to busboy.
It is particularly useful to throttle file uploads, as demonstrated in the test.
I needed this to be able to test my upload progress UI on localhost, but i suppose it could fit other purposes as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ Key | Description
 `fileFilter` | Function to control which files are accepted
 `limits` | Limits of the uploaded data
 `preservePath` | Keep the full path of files instead of just the base name
+`transform` | Pipe request through this stream before busboy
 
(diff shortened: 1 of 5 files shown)


kapouer:
oops, edited to fix syntax error in package.json

kapouer:
It seems the new multer stream API does not cover this use case.
See also https://github.com/expressjs/multer/pull/399#issuecomment-291187075
> what are your thoughts about an option to pipe the request's stream file directly ?
this way it would be possible to request -> throttle -> process multipart -> store.

kapouer:
I managed to throttle incoming request.socket instead, so this PR is useless.

PR #450, opened 2017-01-24
Pass the preservePath down to Busboy

monkbroc:
I'm working on a cloud compiler service that allows uploading files from a directory tree so it needs access to the entire path of each file that a user uploads.

There's a `preservePath` option in Busboy but it's not exposed in Multer so I added an option to Multer to allow preserving paths for uploaded files.

PR #488, opened 2017-05-10
Pass the defCharset down to Busboy

yokomotod:
Hi. This PR  is to enable `defCharset` option in busboy init step,
and handle `multipart/form-data` which contains non-utf8 encoding fields

(related to issue #320 , similar with PR #450 )

Purpose:
I would like to handle `multipart/form-data` which contains various encoding fields.

Now busboy decode field text is utf8 when the field doesn't have charset.
To handle unknown encoding text in later process, it has to be encoded as `'binary'`
so we need to pass `defCharset: 'binary'` to busboy.


UlisesGascon:
Superseded by #1465

PR #491, opened 2017-05-21
Updated README.md to use consts

iddogino:
Best practice for data that shouldn't be changed (at least the pointers)


ctcpip:
already addressed, thank you!

PR #542, opened 2017-11-11
update translation list in the readme.md

deffjoowon:
Your read.md have korean translation. But in the read.md doesn't have that information. So I add that.


deffjoowon:
I have updated. Then it is fixed.

PR #551, opened 2017-12-27
Update README-zh-cn.md

lyh2668:
这个fileFilter在使用的时候会产生一些误解，虽然最后一句的cb描述了传入错误信息的情形，但是如果发生错误了应该是cb(new Error('som errmsg'), false) 如果不添加false则默认返回是成功的


doc/README-zh-cn.md
```diff
@@ -163,95 +163,96 @@ var storage = multer.diskStorage({
...
   cb(null, false)
 
-  // ��������ļ���ʹ��`true`��������:
+  // 接受这个文件，使用`true`，像这样:
   cb(null, true)
 
-  // ��������⣬���������������һ�����:
-  cb(new Error('I don\'t have a clue!'))
+  // 如果有问题，你可以总是这样发送一个错误:
+  cb(new Error('I don\'t have a clue!'), false)
```
LinusU: When an error is passed, the second parameter won't be looked at, so this shouldn't be necessary 🤔

Did this not work for you without the `false` parameter?

LinusU:
Any idea why the diff shows almost every row?

lyh2668:
@LinusU Sorry, when I read the source code, I found it was my own mistake

Issue #568, opened 2018-03-13
[new-api] cause EPERM on windows


Delagen:
How to read from stream?
```
data.file[0].stream
							.on("data", function(chunk) {
								console.log("chunk", chunk);
							})
							.on("end", function() {
								console.log("end");
							});
```
does not work without call `open()` and REMOVING this code
```
file.stream.on('open', function () {
fs.unlink(file.path, function () {})
})
```

LinusU:
`EPERM` indicates that it's a permission issue, does your user have read/write access to the temp directory (`os.tmpdir()`)?

Delagen:
Yes. It remove it without this code when request released.

Delagen:
Sorry. I am wrong. But file does not removed and keep lock if file doesn't read in request. This wrong behavior

Delagen:
Sorry it works. But file does not removed if it doesn't try to read. So I need to add some flush handler

Delagen:
I found issue in my codebase with stream opening, but I open PR to make stream opening and flush more consistent.

Delagen:
@LinusU please review my PR

Delagen:
???

PR #570, opened 2018-03-13
Fix resourse cleaning

delagen:
request stream by demand, fix fs.unlink to call when response send to client or stream open

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: false
 language: node_js
 node_js:
+  - '8'
   - '6'
   - '4'
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -22,10 +22,30 @@ module.exports = function createMiddleware (setup) {
         var appendFile = createFileAppender(options.fileStrategy, req, options.fields)
 
         result.files.forEach(function (file) {
-          file.stream = fs.createReadStream(file.path)
-
-          file.stream.on('open', function () {
+          var flushFn = function () {
             fs.unlink(file.path, function () {})
+            if (res) {
+              res.removeListener('finish', flushFn)
+              res.removeListener('close', flushFn)
+            }
+          }
+
+          if (res) {
+            res.on('finish', flushFn)
+            res.on('close', flushFn)
+          }
+
+          var stream
+
+          Object.defineProperty(file, 'stream', {
+            get: function () {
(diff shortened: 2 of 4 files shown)


Delagen:
For #568

Delagen:
@LinusU It does not fix EPERM error, it's my codebase error.
But it solve unneeded stream opening when code doesn't need it.
And remove files after handler ends and response finished even if files doesn't read in it.

Without this PR files are left opened till process exit if user handler do nothing with them

LinusU:
Hmm, I'm afraid that relying on this could have unintended consequences. For example, I'm not sure that the following code would still work:

```js
app.post('/upload', upload.single('file'), async (req, res) => {
  res.send(200, 'OK')

  const fileName = await allocateFileName()
  const target = fs.createWriteStream()

  req.file.stream.pipe(target)
})
```

Basically, I'm not sure (I don't have a definitive decision, just want to bring up discussion) that not using the stream before responding to the http request means that you don't want to use the file. 🤔

Delagen:
@LinusU this code should work, but it wrong because it can crash, but user does not get any info about it.
I only moved create stream to singleton with getter, and fix removing files at response finish.
But without my PR
```
app.post('/upload', upload.single('file'), async (req, res) => {
  res.send(200, 'OK')
})
```
It will create tmp file and DOES NOT remove it at any time and takes handle, so app may reach open_files limit and crash.

LinusU:
> this code should work

I don't think that it does, since the file will have been removed by the time it hits the `req.file.stream` getter.

> but it wrong because it can crash, but user does not get any info about it

That doesn't necessarily make it _wrong_. There could be plenty of reasons for not wanting to notify the user of errors.

> will create tmp file and DOES NOT remove it at any time and takes handle, so app may reach open_files limit and crash.

The files should be removed, but I just saw that this does in fact leak file descriptors. Basically it boils down to the underlying FD not beeing freed when the `ReadStream` is garbage collected.

This is potentially something that should be fixed in Node.js, I'll open an issue about that.

We could probably use e.g. [node-weak](https://github.com/TooTallNate/node-weak) to work around that manually for now.

noonii:
+1

I noticed in `multer@next` that OP is right it does not get deleted from temp file upon failed or successful request.

When I re-run the app, only then it is removed.

PR #575, opened 2018-04-05
Avoid Stale when Callback Called Before Busboy Finished

dharijanto:
Earlier, when StorageEngine's 'cb' callback of _handleFile()
is called before busboy finished its processing, the callback
to the client side is never called.


LinusU:
`indicateDone` is being called both when all the files have been written, and when busboy is done with the stream.

The idea is that it will wait for _both of them_ before calling `done`.

What is the problem that you are experiencing?

Issue #579, opened 2018-04-11
documentation addition


tsando:
Would be nice if you could add to the documentation how your module links to the HTML forms. I posted an example here https://stackoverflow.com/a/49773878/4303016. We can easily add this to the examples part in the documentation?

LinusU:
PR welcome 🙌

tsando:
@LinusU done, you just need to accept my PR

PR #580, opened 2018-04-17
updated readme with example on how to link to html multipart form

tsando:
This addresses https://github.com/expressjs/multer/issues/579#issuecomment-381947094

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Here's an example on how multer is used an HTML form. Take special note of the `
 </form>
 ```
 
-Then in your javascript file you would add these lines to access both the file and the body:
+Then in your javascript file you would add these lines to access both the file and the body. It is important that you use the `name` field value from the form in your upload function. This tells multer which field on the request it should look for the files in. If these fields aren't the same in the HTML form and on your server, your upload will fail:
 
 ```javascript
 var multer  = require('multer')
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,6 +66,32 @@ app.post('/profile', upload.array(), function (req, res, next) {
 })
 ```
 
+Here's an example on how this links to an HTML form. Suppose we have a form group with enctype="multipart/form-data" as follows:
```
```suggestion
Here's an example on how multer is used an HTML form. Take special note of the `enctype="multipart/form-data"` and `name="uploaded_file"` fields:
```

README.md
```diff
@@ -66,6 +66,32 @@ app.post('/profile', upload.array(), function (req, res, next) {
...
+<form action="/stats" enctype="multipart/form-data" method="post">
+  <div class="form-group">
+    <input type="file" class="form-control-file" name="uploaded_file">
+    <input type="text" class="form-control" placeholder="Number of speakers" name="nspeakers">
+    <input type="submit" value="Get me the stats!" class="btn btn-default">            
+  </div>
+</form>
+```
+
+Then in your javascript file you would add these lines to access both the file and the body:
```
```suggestion
Then in your javascript file you would add these lines to access both the file and the body. It is important that you use the `name` field value from the form in your upload function. This tells multer which field on the request it should look for the files in. If these fields aren't the same in the HTML form and on your server, your upload will fail:
```

Issue #607, opened 2018-06-27
Filesize in mutliple upload returns fieldname instead of filename


cKhoff:
that is worth when you want to upload multiple files and one file is above filesize limitation.
you´re not able to inform user about that rejected file...
https://github.com/expressjs/multer/blob/master/lib/make-middleware.js#L141 pls, change fieldname to filename, or even better with the file object

LinusU:
could be nice to add the `filename` as well 👍

cKhoff:
here´s my PR :)
https://github.com/expressjs/multer/pull/608

cKhoff:
https://github.com/expressjs/multer/pull/611 might could be implemented soon ?

UjjwalKumar239:
I'd like to work on this issue. 
I'll update the error object to include the filename 
so users can identify which file exceeded the size limit.

PR #608, opened 2018-06-27
exceeded filesize Limit return value corrected to upper file Object

cKhoff:
in addition to https://github.com/expressjs/multer/issues/607
just changed fieldname to the file object, so that u have the choice, to do own errorhandling on exceeded filesize error.

This is very useful when you want to upload many files and don´t want to reject the hole array.
now You can inform the User which file was to large to upload, without rejecting the hole upload request.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -138,7 +138,7 @@ function makeMiddleware (setup) {
 
         fileStream.on('limit', function () {
           aborting = true
-          abortWithCode('LIMIT_FILE_SIZE', file)
+          abortWithCode('LIMIT_FILE_SIZE', fieldname, file)
         })
 
         storage._handleFile(req, file, function (err, info) {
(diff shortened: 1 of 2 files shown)


LinusU:
While this would be nice, this would be a breaking change. In order to get this into Multer 1.x we have to keep the current functionality working as is.

lib/make-middleware.js
```diff
@@ -138,7 +138,7 @@ function makeMiddleware (setup) {
 
         fileStream.on('limit', function () {
           aborting = true
-          abortWithCode('LIMIT_FILE_SIZE', fieldname)
+          abortWithCode('LIMIT_FILE_SIZE', file)
```
LinusU: Can you change so that we can call this like so:

```js
abortWithCode('LIMIT_FILE_SIZE', fieldname, file)
```

cKhoff:
so you think https://github.com/expressjs/multer/pull/608/commits/2d83fb1af91fe4edf32b92a65fded4108af5df26  might be backward compatible ?

PR #609, opened 2018-06-28
Bump dependencies

LinusU:


package.json
```diff
@@ -19,22 +19,22 @@
     "middleware"
   ],
   "dependencies": {
-    "append-field": "^0.1.0",
+    "append-field": "^1.0.0",
```
LinusU: Bump was just to mark as stable

package.json
```diff
@@ -19,22 +19,22 @@
...
     "busboy": "^0.2.11",
     "concat-stream": "^1.5.2",
     "mkdirp": "^0.5.1",
-    "object-assign": "^3.0.0",
+    "object-assign": "^4.1.1",
```
LinusU: Only breaking change was changing internal ordering that doesn't affect us

package.json
```diff
@@ -19,22 +19,22 @@
...
     "on-finished": "^2.3.0",
     "type-is": "^1.6.4",
     "xtend": "^4.0.0"
   },
   "devDependencies": {
     "express": "^4.13.1",
     "form-data": "^1.0.0-rc1",
-    "fs-temp": "^0.1.2",
-    "mocha": "^2.2.5",
+    "fs-temp": "^1.1.2",
```
LinusU: Stricter validation on input, we already passed

package.json
```diff
@@ -19,22 +19,22 @@
...
+    "mocha": "^3.5.3",
```
LinusU: Dropped support for Node.js 0.8

package.json
```diff
@@ -19,22 +19,22 @@
...
     "rimraf": "^2.4.1",
-    "standard": "^8.2.0",
+    "standard": "^11.0.1",
```
LinusU: New linting rules, we already pass

PR #611, opened 2018-06-29
added file object to exceeded FileSizeLimit Error response

cKhoff:


lib/make-middleware.js
```diff
@@ -138,7 +138,7 @@ function makeMiddleware (setup) {
 
         fileStream.on('limit', function () {
           aborting = true
-          abortWithCode('LIMIT_FILE_SIZE', fieldname)
+          abortWithCode('LIMIT_FILE_SIZE', fieldname, file)
```
LinusU: This won't do anything since `abortWithCode` only takes two arguments

cKhoff: i´ll try to alter it when i got some spare time

PR #625, opened 2018-08-04
Added fs.exists for dest to check if the directory as already created

laynef:
I currently can not use this package with the option: dest with a multithreaded server in Node.js
It will cause my server to crash because it has already been created.

I added a if statement for if the directory path already exists


laynef:
I discovered the problem I was having. Separate issue. You guys rock, I love your code base. Next time I'd provide documentation and testings if you need any help or if I discover anything I'll test with isolated testing.

PR #641, opened 2018-08-26
Make Multer errors inherit from MulterError

turakvlad:
For now, it's hard to indicate whether an error thrown by the Multer package was the Multer's one or not. Let's take this sample as an example:

```
var upload = multer().single('avatar')

app.post('/profile', function (req, res) {
  upload(req, res, function (err) {
    if (err) {
      // An error occurred when uploading
      return
    }

    // Everything went fine
  })
})
```

An `err` may be on of the [Multer's errors](https://github.com/expressjs/multer/blob/master/lib/make-error.js#L1-L9) or one of the [common system errors](https://nodejs.org/api/errors.html#errors_common_system_errors). For example, if I am trying to save a file to a directory that does not exist, I will get `ENOENT` error. For example, these functions may throw common system errors:

https://github.com/expressjs/multer/blob/ee5188d7499dd19c8596fe011e6f6e53ea4778d6/lib/make-middleware.js#L117
https://github.com/expressjs/multer/blob/ee5188d7499dd19c8596fe011e6f6e53ea4778d6/lib/make-middleware.js#L136
https://github.com/expressjs/multer/blob/ee5188d7499dd19c8596fe011e6f6e53ea4778d6/lib/make-middleware.js#L154
https://github.com/expressjs/multer/blob/ee5188d7499dd19c8596fe011e6f6e53ea4778d6/lib/make-middleware.js#L167

What if I want to return `400 Bad Request` in the case of the Multer's error and pass the common system errors to the next error handler middleware with `next(err)`?

For now, I can do this with:

```
if ([
  'LIMIT_PART_COUNT', 
  'LIMIT_FILE_SIZE', 
  'LIMIT_FILE_COUNT', 
  'LIMIT_FIELD_KEY', 
  'LIMIT_FIELD_VALUE', 
  'LIMIT_FIELD_COUNT', 
  'LIMIT_UNEXPECTED_FILE'
].includes(err.code)) {
  console.log(`Multer's error here! I can return 400 Bad Request!`)
} else {
  console.log(`Unkown error here! It looks like I need to pass it to the next error handler middleware`)
}
```

It doesn't look nice for me. On my opinion, these methods will be better:

```
if (err instanceof MulterError) { console.log(`Multer's error here! I can return 400 Bad Request!`) }
```

or

```
if (err.isMulter) { console.log(`Multer's error here! I can return 400 Bad Request!`) }
```

The first option looks nice but I am not sure whether it will be a breaking change or not. In this case we need `MulterError` class to extend `Error` class and all the errors should extend `MulterError` class instead of `Error` class (please, look [here](https://github.com/expressjs/multer/blob/master/lib/make-error.js#L12)). 

The second option works great too and I think that it won't be a breaking change. For example, [Boom](https://github.com/hapijs/boom#boom) and [Joi](https://github.com/hapijs/joi/blob/v13.6.0/API.md#errors) are working the same way.

What do you think about it? It it looks good for you, I can add this information to the `README.md`.


LinusU:
Personally prefer extending `MulterError` from `Error` I think 🤔

turakvlad:
@LinusU, in this case, I should create a file called `./lib/MulterError.js` which basically will be ES5 function constructor. It should look like:

```javascript
module.exports = function MulterError(message, code, field) {
  Error.captureStackTrace(this, this.constructor);
  this.name = this.constructor.name;
  this.message = message;
  this.code = code;
  this.field = field;
};

require('util').inherits(module.exports, Error);
```

I have used [this](https://gist.github.com/justmoon/15511f92e5216fa2624b) as an inspiration. Without `require('util').inherits(...)` it would be quite hard to extend the `Error` class. Please, refer to [this](https://stackoverflow.com/questions/783818/how-do-i-create-a-custom-error-in-javascript) and [this](https://stackoverflow.com/questions/8802845/inheriting-from-the-error-object-where-is-the-message-property) questions on the Stackoverflow. I prefer `require('util').inherits(...)` rather than using those wierds "hacks" to extend the `Error` class.

After that this new function constructor should be required into `./lib/make-error.js` file. The file should look like:

```javascript
var MulterError = require('./MulterError')

var errorMessages = {
  'LIMIT_PART_COUNT': 'Too many parts',
  'LIMIT_FILE_SIZE': 'File too large',
  'LIMIT_FILE_COUNT': 'Too many files',
  'LIMIT_FIELD_KEY': 'Field name too long',
  'LIMIT_FIELD_VALUE': 'Field value too long',
  'LIMIT_FIELD_COUNT': 'Too many fields',
  'LIMIT_UNEXPECTED_FILE': 'Unexpected field'
}

function makeError (code, field) {
  return new MulterError(errorMessages[code], code, field)
}

module.exports = makeError
```

Also, `.lib/MulterError` should be required to the main `index.js` file and it should be exported at the end of the file the same way as `diskStorage` and `memoryStorage`:

```javascript
var diskStorage = require('./storage/disk')
var memoryStorage = require('./storage/memory')
var MulterError = require('./lib/MulterError') // This line was added.

// A lot of code goes here...

module.exports = multer
module.exports.diskStorage = diskStorage
module.exports.memoryStorage = memoryStorage
module.exports.MulterError = MulterError // This line was added.
```

After these changes, we will be able to do in our code:

```javascript
var multer = require('multer')
var upload = multer().single('avatar')

app.post('/profile', function (req, res) {
  upload(req, res, function (err) {
    if (err instanceof multer.MulterError) {
      console.log(`Multer's error here! I can return 400 Bad Request!`)
      console.log(err.name); // MulterError
      console.log(err.message); // Unexpected field
      console.log(err.field); // avatar
      console.log(err.code); // LIMIT_UNEXPECTED_FILE
      console.log(err.stack); // Stack goes here...
      console.log(err instanceof Error); // true
      console.log(err instanceof multer.MulterError); // true
    } else {
      console.log(`Unkown error here! It looks like I need to pass it to the next error handler middleware`)
    }

    // Everything went fine
  })
})
```

Also, in the `.tests/error-handling` I should add one more test that checks whether `err` is MulterError's instance.

All the current tests are passing successfully. Does it look good for you? If yes, I can commit these changes and change `README.md`. If no, please let me know what you want to change / add.

Thanks.

LinusU:
Looks great 👍 

Just some nits:

```js
var util = require('util')

function MulterError (message, code, field) {
  Error.captureStackTrace(this, this.constructor)
  this.name = this.constructor.name
  this.message = message
  this.code = code
  this.field = field
}

util.inherits(MulterError, Error)

module.exports = MulterError
```

and can you name the file `mutler-error.js` to be consistent with the others? ☺️

turakvlad:
@LinusU, hey. Please, take a look at this PR when you have time. If you have any suggestions, please let me know.

lib/multer-error.js
```diff
@@ -0,0 +1,13 @@
+var util = require('util')
+
+function MulterError (message, code, field) {
+  Error.captureStackTrace(this, this.constructor)
+  this.name = this.constructor.name
+  this.message = message
+  this.code = code
+  this.field = field
```
LinusU: Could you add back the `if` statement? As it is now, `'field' in err` will evaluate to true, even for errors without fields.

turakvlad: Ugh, I am sorry. I didn't notice that.

lib/make-error.js
```diff
@@ -1,3 +1,5 @@
+var MulterError = require('./multer-error')
+
 var errorMessages = {
```
LinusU: Could we move this map to the `multer-error.js` file, remove `message` as a constructor argument to `MutlerError`, replace all calls to `makeError` with `new MulterError`, and finally remove this file? Thanks ☺️

turakvlad: Great idea! Just did it. Please, take a look.

turakvlad:
@LinusU, hey. Could you please take a look at the new changes?

turakvlad:
@LinusU, hi. You still don't merge my PR.

Please, let me know if something is wrong with the code. I will fix it asap.

LinusU:
Awesome work, thanks 👏

LinusU:
Published as 1.4.0 🚀

Issue #634, opened 2018-08-18
Question about .none() method


turakvlad:
In the `README.md` I can see:

> #### `.none()`
> Accept only text fields. If any file upload is made, error with code
> "LIMIT\_UNEXPECTED\_FILE" will be issued. This is the same as doing `upload.fields([])`.

As I understand .`none()` is the same as doing not only `upload.fields([])`, but the `.single()` and `.array()` too? Am I correct?

The next methods do the same things?

- `.none()`
- `.single()`
-  `.array()` 
- `upload.fields([])`

If I am correct, I think it'd be better to remove this from the `REAME.md`:

> In case you need to handle a text-only multipart form, you can use any of the multer methods (`.single()`, `.array()`, `fields()`). Here is an example using `.array()`:
> ```javascript
> var express = require('express')
> var app = express()
> var multer  = require('multer')
> var upload = multer()
> 
> app.post('/profile', upload.array(), function (req, res, next) {
>   // req.body contains the text fields
> })
> ```

In my opinion, it is better not to confuse the readers with a few methods that are doing the same thing. It would be better to show only one right way - `.none()`.

Thanks.

LinusU:
It's not the same, you shouldn't call `single`, `array` or `fields` without any arguments. That should throw an error, but probably doesn't at the moment.

You can see the different methods here: https://github.com/expressjs/multer/blob/ee5188d7499dd19c8596fe011e6f6e53ea4778d6/index.js#L59-L73

As you can see, calling e.g. `single()` will actually result in `[{ name: undefined, maxCount: 1 }]`.

Calling `none()` will also pass the type `'NONE'` which makes it so that neither `req.file` or `req.files` gets set. Therefore I'd be happy to accept a PR that removes this sentence, since it's not quite true:

> This is the same as doing upload.fields([]).

turakvlad:
@LinusU, thanks for the answer!

What about this section:

> In case you need to handle a text-only multipart form, you can use any of the multer methods (`.single()`, `.array()`, `fields()`). Here is an example using `.array()`:
> ```javascript
> var express = require('express')
> var app = express()
> var multer  = require('multer')
> var upload = multer()
> 
> app.post('/profile', upload.array(), function (req, res, next) {
>   // req.body contains the text fields
> })
> ```

Should I remove it too? `upload.array()` is invoked with no arguments here.

LinusU:
I think that we should only advocate using `.none()` for handling value-only multipart forms 👍

[...]

PR #644, opened 2018-08-27
Update README.md

turakvlad:
Please, refer to https://github.com/expressjs/multer/issues/634.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ app.post('/cool-profile', cpUpload, function (req, res, next) {
 })
 ```
 
-In case you need to handle a text-only multipart form, you can use any of the multer methods (`.single()`, `.array()`, `fields()`). Here is an example using `.array()`:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,15 +53,15 @@ app.post('/cool-profile', cpUpload, function (req, res, next) {
 })
 ```
 
-In case you need to handle a text-only multipart form, you can use any of the multer methods (`.single()`, `.array()`, `fields()`). Here is an example using `.array()`:
+In case you need to handle a text-only multipart form, you should use `none()` method:
```
LinusU: English is not my first language, but I think that there should be a "the" in there. Also, could you add a `.` before `none()`?

> In case you need to handle a text-only multipart form, you should use the `.none()` method:

LinusU:
Nice! 👏 

Just one small nit...

turakvlad:
@LinusU, I've fixed the typos :smile:

LinusU:
Thanks!

PR #658, opened 2018-10-02
(not sure) Fix a logical branchment in the readme

Ostefanini:
I am not sure, but it seems better with this, to avoid entering in the wrong branchment (`an unknown error occured when uploading`) when everything wents fine ?


LinusU:
Thanks!

PR #669, opened 2018-10-12
tests: remove old Node.js versions

DanielRuf:
All versions prior Node.js 6 reached their EOL.

https://github.com/nodejs/Release/blob/master/README.md

If we increase the min version in package.json we should do this as part of a major release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - "0.10"
-  - "0.12"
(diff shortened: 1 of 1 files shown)


LinusU:
This is already done in the 2.x-alpha, we will still support Node.js 0.10 for all 1.x releases, and must thus still test for them ☺️

DanielRuf:
Hm, why not make the 2.x branch the current default branch for us contributors?

PR #671, opened 2018-10-14
Add option "mkdirOpts" which will pass into mkdirp

PuHsiu:
Make unprivileged user access the directory created when diskStorage instance created possible (pass "mode" option into mkdirp).


UlisesGascon:
Thanks @PuHsiu, and sorry for the wait. We're not adding an option for this: creating the folder when `destination` is a string is only a convenience for the simple case. If you need specific permissions, create the directory yourself before calling `multer()`, for example `fs.mkdirSync(dir, { recursive: true, mode: 0o775 })`, and multer will just use it. `mkdirp` is also no longer a dependency.

PR #693, opened 2018-12-07
allow file upload without filename

warncke:
This is an edge case where a client is sending an upload request without a filename but with the content type of application/octet-stream so it gets silently ignored.

Per the spec filename is not required (https://tools.ietf.org/html/rfc1867) so even if this is a rare case it seems reasonable to allow.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -94,8 +94,10 @@ function makeMiddleware (setup) {
 
     // handle files
     busboy.on('file', function (fieldname, fileStream, filename, encoding, mimetype) {
-      // don't attach to the files object, if there is no file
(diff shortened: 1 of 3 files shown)


LinusU:
Hmmm, do you think that this should be a breaking change? 🤔 

I'm guessing yes maybe? 🤔

warncke:
i don't have a good answer on this. it does change the behavior significantly, but it applies to a very narrow set of cases, and it makes the behavior more consistent with the specs, so not sure how you classify it.

at the end of the day i don't think many people are encountering this issue or it would have been raised before so probably best to err on the side of caution and not breaking anything for existing users.

Issue #39, opened 2014-06-18
Dynamically changing destination folder name


kokujin:
Is it possible to dynamically change the destination folder name depending on the field name? I made a quick test which failed

```
app.use(multer({
  dest: './uploads/',
  rename: function (fieldname, filename) {
      if(fieldname == 'avatar_file'){
          return filename.replace(/\W+/g, '-').toLowerCase() + Date.now()
      } else {
          return 'avatars/' + filename.replace(/\W+/g, '-').toLowerCase() + Date.now()
      }
  }
}))
```

Thanks

Mindstormer619:
@kokujin If you are still interested, I managed to do this with this small addition:

```js
"rename" : function (fieldname, filename, req, res) {
	return path.join(fieldname, "i" + req.params.user_id);
}
```

(Saves it like `<defaultFolder>/<fieldName>/i564221sdd456.jpg`. I added the "i" because some stuff might complain with filenames starting with numbers)

I got the idea after looking at multer's source code. It seems that it does a `path.join()` to get the full path using the destination folder and the uploaded filename. Adding this works for me.

Note that this might change in the future, breaking this code? @hacksparrow?

Mindstormer619:
Actually is there a specific reason why fieldnames and filenames cannot be passed to `changeDest()`?

LinusU:
@Mindstormer619 I think that the problem with using fieldnames is that they might not be available when the file comes. Most browsers sends the fields and files in the order they are placed in the DOM. So if the field for the name is after the file input field, multer won't know about it when it's receiving the file.

Mindstormer619:
Ohhh, interesting. So I'm taking it that the `rename` feature works after the file is done uploading?

Mindstormer619:
Hmm, looking at the source it doesn't seem to be a problem to be including the field name as a param for the `changeDest()` function. I could be wrong about this. Can someone confirm?

Mindstormer619:
@LinusU I'm fairly sure that multer is well aware of the fieldname before the file is uploaded, as can be referenced by `file.fieldname` within the `onFileUploadStart` function

LinusU:
@Mindstormer619 Ahh, I think I misunderstood. I thought the question was about renaming it to the value of another field. The fieldname is known to multer at that point, although not passed to the `changeDest` function.

Your workaround should work :+1:

Mindstormer619:
[...]

PR #721, opened 2019-03-06
Append a note about chaning desination folers in multer

AndrewKralovec:
Adding my workaround to the readme as mentioned in https://github.com/expressjs/multer/issues/39#issuecomment-470050186


UlisesGascon:
Thanks @AndrewKralovec! The README already covers this... `destination` can be a `function (req, file, cb)`, which is the supported way to pick a folder per upload. So I'm closing this one.

Issue #690, opened 2018-11-30
MulterError: Unexpected field


zoukie33:
Hello, 
i've tried to upload a file with postman, but i get an "MulterError: Unexpected field", i search in all issues and i see all the time this is the field name. But for me i put the good filedname in postman and in my code.

here my code (sorry for screens, my first time posting code in issues):

![image](https://user-images.githubusercontent.com/5038862/49307669-2d994c00-f4d6-11e8-8a12-c4c691da3566.png)

here my postman request:
![shitappend](https://user-images.githubusercontent.com/5038862/49307299-05f5b400-f4d5-11e8-8c3b-f588192d9b74.jpg)

Thank your for the help.

zoukie33:
hello again,
i've done what you told me, and i get: "This is the invalid field -> pic".
but the file field in postman is goo and in the code too.

![field_error](https://user-images.githubusercontent.com/5038862/49387863-25821c00-f723-11e8-8e83-97a9eb502008.jpg)

LinusU:
Hmm, interesting 🤔 

Could it be that there is more than one muter attached as a middleware somehow? It's hard for me to tell since I can't see the entire code...

That's the only thing I can think of, otherwise, could you try and reduce this to the absolute minimal code that's required to show this behaviour and upload somewhere, and then I could take a look at it 👀

zoukie33:
> Could it be that there is more than one muter attached as a middleware somehow? It's hard for me to tell since I can't see the entire code...

hum i dont really know, but i have jwt protection for this routes, and i have an invalid Access if i dont put this in my app.js :
`app.use(upload.array());
app.use(express.static('public'));
app.use('/files', express.static('files'));`

I've tried without upload.array(), not working and with .none() the same.
this in my app.js can be the problem, but without that i can't access to the route.

for now i just uploaded the 3 files concerned about this problem, if you want i can make you a working version without all my routes (because if you run my api without everything behind this won't work ^^).

here the files
https://ufile.io/c3yp2

my app is working on ubuntu server:  Ubuntu 18.04.1 LTS

LinusU:
```js
app.use(upload.array());
```

This line is probably the problem, it shouldn't be there as it will catch all files uploaded to any route and then complain that that they have the wrong field (since none was supplied to that function).

zoukie33:
ok, so what can i do to fix that ? cause if i dont put it in my add.js, et post route who need an upload cant be reach if this is not here ... 
[...]

PR #732, opened 2019-04-12
Show field name in "Unexpected field" error

hubgit:
To help with debugging "Unexpected field" errors, show which field was unexpected.


hubgit:
As requested in https://github.com/expressjs/multer/issues/690#issuecomment-443749868

Issue #455, opened 2017-02-02
Should allow absent/empty filename


LinusU:
As far as I can tell from reading [RFC7578 Section 4.2](https://tools.ietf.org/html/rfc7578#section-4.2) the filename is **not** mandatory. Currently multer ignores files without a filename, [relevant code here](https://github.com/expressjs/multer/blob/7ef2e81215e9b773204df44fb4aee29e58340061/lib/make-middleware.js#L97-L98).

I think that this should be addressed, though I'm not sure if we should set `originalname` to an empty string or to `undefined`...

This might possibly be a breaking change as well, I'll start by fixing it in the 2.x beta...

LinusU:
Fixed in `v2.0.0-alpha.5`

Ahimta:
I would very much like to use multer but I can't right now because some clients upload files without filename. When will the final version be released :sweat_smile:?

LinusU:
I would recommend using the alpha versions, they should be stable and I know a few companies that uses them in production ☺️

jonbarrow:
Is using the alpha versions still considered stable enough for production use? It's been 5 years since that was said, so I wanted to ask to be safe. And is there any news on this making it to a stable release at some point? We have clients who almost never send file names which prevents us from using multer and instead use our own parser, but that becomes a bit of a hassle since we need to implement this on multiple services

bjohansebas:
Version 2 is the LTS version of version 1. To get this change again, you need to upgrade to the alpha of version 3, which includes the changes you need.

https://expressjs.com/2025/05/19/security-releases.html

PR #735, opened 2019-04-19
feat(form-data): Adds support for parsing regular blobs / non-file data (e.g. application/json)

simllll:
this change allows parsing files without "filenames".. which is the case if you insert a blob with a content-type json/application for example:

```
const form = new FormData();
form.append('jsonStuff', new Blob([JSON.stringify({a: 1})], {type: 'application/json'}, ''); // passing a 3rd empty paramter sets the filename to an empty string, if not, the default value is "blob". Setting parseBlobs to true, would also handle the filename "blob" in this case.
```

results into:
------WebKitFormBoundaryeFyAvRTAPR4MqtxM
Content-Disposition: form-data; name="jsonStuff"; filename=""
Content-Type: application/json

{a: 1}

Therefore multer will be able to handle non-files and it is possible to send files and json data within one request.

This topic has a history in busboy, see mscdex/busboy#200 and mscdex/busboy#201

But I agree, this should be handled within multer.

It's a non breaking change, as empty filenames have been ignored till today anyway, and the "blob" filename case is just a new option that needs to be set explicty.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -3,6 +3,7 @@ var Busboy = require('busboy')
 var extend = require('xtend')
 var onFinished = require('on-finished')
 var appendField = require('append-field')
+var concat = require('concat-stream')
 
 var Counter = require('./counter')
 var MulterError = require('./multer-error')
@@ -24,6 +25,7 @@ function makeMiddleware (setup) {
     var fileFilter = options.fileFilter
     var fileStrategy = options.fileStrategy
     var preservePath = options.preservePath
+    var parseBlobs = options.parseBlobs
 
     req.body = Object.create(null)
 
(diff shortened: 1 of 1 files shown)


simllll:
This topic has a history in busbox, see https://github.com/mscdex/busboy/pull/200 and https://github.com/mscdex/busboy/issues/201 

But I agree, this should be handled within multer.

It's a non breaking change, as empty filenames have been ignored till today anyway, and the "blob" filename case is just a new option that needs to be set explicty.

simllll:
any comment on this PR?

mlakmal:
@simllll i am having same issue on our api. do you have anyother option to override multer until this PR is merged?

simllll:
Hi @mlakmal , I just pushed a fork including the fix to our npm namespace @hokify/multer

https://www.npmjs.com/package/@hokify/multer

let me know if this works for you

regards
Simon

mlakmal:
@simllll thanks for the fork. This should work for us (just need to get internal approval). I’ll also watch this issue for any updates on original package.

mlakmal:
@simllll looks like parseBlobs option is not getting passed to makeMiddleware function on your fork. would you be able to fix that ?

simllll:
Just published 1.5.1 .. please retry. should work now, I'm not using the parseblob option, therefore I never really tested it.

Please let me know if it works now

Regards

simllll:
Any news about this PR?

simllll:
Any comment on this PR @LinusU ?

simllll:
Hello? @LinusU?

UlisesGascon:
Thanks @simllll, and sorry this never got an answer. We're not going to add content parsing to multer: it hands multipart parts to `req.body` or to a storage engine, and decoding JSON or CSV payloads belongs to the application. The underlying gap, parts without a filename being skipped, is tracked in #455 and handled in the v3, where such a part arrives as a file and `JSON.parse(file.buffer)` is a one-liner :thinking:

Issue #627, opened 2018-08-07
Hangs on jpeg upload


E-Rockalanche:
I'm able to get multer to upload a png image just fine, but when I try to upload jpeg images of varying sizes they all make multer hang. They pass the file size limit and fileFilter function. They all get saved into uploads/temp/. Multer just doesn't go to the next middleware!

Here's an example that fails for me:

```
// route.js

const filesList = [
  {
    name: 'selfie',
    maxCount: 1,
  },
];

const maxFileSize = 2 * 1024 * 1024;

const upload = multer({
  dest: 'uploads/temp',
  limits: {
    fileSize: maxFileSize,
  },
  
  fileFilter: (req, file, cb) => {
    console.log('checking file types');
    const supportedMimeTypes = [
      'image/jpeg',
      'image/png',
    ];
    if (!supportedMimeTypes.includes(file.mimetype)) {
      console.log('file type not supported');
      cb(null, false);
      return;
    }
    
    console.log('file: ', file);
    
    console.log('file ok');

    // all images reach this far

    cb(null, true);
  },
}).fields(filesList);

const middleware = (req, res, next) => {
  console.log('Parsing multipart');
  upload(req, res, (err) => {

    // jpeg images never reach this far, but png images do

    console.log('callback');
    if (err) {
      console.log('Error from multer: ', err);
      throw err;
    }
    next();
  });
};
```

E-Rockalanche:
I don't think it's a jpeg problem anymore. Multer also hangs when I send multiple pngs

E-Rockalanche:
I am using chai-http to test my server if that makes a difference.

```
it.only('Should upload files', async function() {
    this.timeout(0);
    
    const body = {
      token,
      firstName: 'Danny',
      lastName: 'Avidan',
      address: '1234 road drive',
      postal: '123456',
      city: 'The Couve',
      province: 'BC',
      country: 'Canada',
    };
    
    const res = await chai.request(app)
      .post('/verification')
      .attach('selfie', fse.readFileSync(__dirname+'/../../../data/selfie.png'), 'selfie.png')
      .attach('bank', fse.readFileSync(__dirname+'/../../../data/bank.jpg'), 'bank.jpg')
      .field(body);
    
    expect(res.status).to.be.equal(201);
    
    const docs = await Document.query();
    expect(docs.length).to.be.equal(2);
    docs.foreach(doc => {
[...]

PR #736, opened 2019-04-20
upgrade busboy to prevent hanging when uploading files

grandfoosier:
see: https://github.com/expressjs/multer/issues/627

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   ],
   "dependencies": {
     "append-field": "^1.0.0",
-    "busboy": "^0.2.11",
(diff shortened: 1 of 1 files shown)


LinusU:
Unfortunately, this breaks compatibility with older versions of Node.js and thus have to be a breaking change. I'll fix this on the v2.x branch 👌

LinusU:
Fixed in 2.x and published as 2.0.0-alpha.7 🚀

ctcpip:
busboy has been upgraded

PR #740, opened 2019-04-24
Fixed Typo

JayTailor45:
Request to fix typo


LinusU:
This was done intentionally as the "optional part" includes the `, `, that is you can write either `.array(fieldname)` or `.array(fieldname, maxCount)`. Does that make sense?

JayTailor45:
@LinusU Gotcha, You are totally correct. It's my bad. :P

PR #751, opened 2019-07-02
chore: Update mocha

Trott:
Update mocha from 3.5 to 4.1. This resolves some issues reported by `npm
audit`.

To work with mocha 6, the express test had to explicitly the server.
Mocha no longer forces an exit when it thinks the test is done, so the
server was keeping the process running, causing the test suite to never
finish.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "express": "^4.13.1",
     "form-data": "^1.0.0-rc1",
     "fs-temp": "^1.1.2",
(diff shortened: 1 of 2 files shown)


Trott:
Moved it back to mocha 4.1 in the hopes of being compatible with Node.js 0.10.

Trott:
Looks like this would have to be a breaking change and the update is already in the `explore-new-api` branch (which appears to be the branch for the 2.x release). Closing.

PR #752, opened 2019-07-05
chore: Update testdata-w3c-json-form

Trott:
Version 1.0.0 of testdata-w3c-json-form contains a correction to the
test data. Without the correction (of the bottle-on-wall type from an
array of numbers to an array of strings), it is possible that tests will
break in Node.js 13.0.0 or a future release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "mocha": "^3.5.3",
     "rimraf": "^2.4.1",
     "standard": "^11.0.1",
-    "testdata-w3c-json-form": "^0.2.0"
+    "testdata-w3c-json-form": "^1.0.0"
   },
   "engines": {
     "node": ">= 0.10.0"


Trott:
(By the way, the "may break in a future Node.js version" comes from https://github.com/nodejs/node/pull/28011. There's an excellent chance that change won't land, but there's a non-zero chance it will, and I hope it will, so... this will make sure it doesn't break multer tests.)

Trott:
Travis is green even the widget says yellow/in-progress. 
¯\\_(ツ)_/¯

LinusU:
> Travis is green even the widget says yellow/in-progress.

Tried restarting one of the builds...

LinusU:
Thanks!

Trott:
Any chance of a patch release getting pushed with this change? I'd like to get https://github.com/nodejs/node/pull/28011 landed in Node.js but multer tests will fail until this change is in a release, and multer is one of the npm modules that gets tested in the tests for breaking changes in Node.js core.

LinusU:
Hmm, I usually don't do patch releases with just a bumped dev-dependency since it wouldn't change anything for the consumers of the library, but sure thing 👍

LinusU:
`multer@1.4.2` published :shipit:

PR #753, opened 2019-07-08
chore: Update mocha from 3.x to 4.x

Trott:
Update mocha from 3.x to 4.x. This removes a pair of warnings from `npm
audit`. The epxress test needed to explicitly close the server when done
as mocha 4.x does not do this by default the way 3.x did.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "express": "^4.13.1",
     "form-data": "^1.0.0-rc1",
     "fs-temp": "^1.1.2",
-    "mocha": "^3.5.3",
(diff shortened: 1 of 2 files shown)


Trott:
mocha 4.x is incompatible with Node.js 0.10 and 0.12 so tests fail with Node.js < 4.x. As much as I'd love to see only currently-supported versions of Node.js used (currently 8.x, 10.x, and 12.x), I imagine that's not in the cards here, so I'll close this.

LinusU:
This will be fixed with #399, can't break compatibility on the 1.x release line unfortunately...

PR #766, opened 2019-09-02
Update readme uploadany

ctrl-alt-lulz:


UlisesGascon:
Thanks @ctrl-alt-lulz! We keep `.any()` out of the quick-start examples on purpose and reminds documented in the API section together with the warning about accepting arbitrary files, and we'd rather people reach for `.single()`/`.array()`/`.fields()` first. Closing this one.

Issue #509, opened 2017-07-13
Usage of deprecated function crypto.pseudoRandomBytes


ghost:
`crypto.pseudoRandomBytes` is deprecated and now the preferred function is `crypto.randomBytes`.

PR #774, opened 2019-09-17
storage: replace deprecated pseudoRandomBytes with randomBytes

gireeshpunathil:
Fixes: https://github.com/expressjs/multer/issues/509
Fixes: https://github.com/expressjs/multer/issues/811

cc @LinusU

### DIFF
--- a/storage/disk.js
+++ b/storage/disk.js
@@ -5,7 +5,7 @@ var crypto = require('crypto')
 var mkdirp = require('mkdirp')
 
 function getFilename (req, file, cb) {
-  crypto.pseudoRandomBytes(16, function (err, raw) {
+  crypto.randomBytes(16, function (err, raw) {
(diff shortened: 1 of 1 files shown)


gireeshpunathil:
bump. Is there anything I can help with, to move this forward? two issues waiting on this. thanks!

geoynomous:
it might be delayed, as older versions of node do not have that function?

gireeshpunathil:
that depends; on :
 - which version of multer this is going to be cut into
 - which version of node `that` version multer is going to support

LinusU:
Sorry for not getting to this!

`crypto.randomBytes` is supported on Node.js 0.10 which is what Multer 1.x is supporting, so there should be no problem merging this 👍

geoynomous:
Is there also a 2.0 release coming?

Issue #238, opened 2015-10-03
Race condition in test


LinusU:
This seems to be very very rare, found it on Travis today.

``` text
$ npm test

> multer@1.0.5 test /home/travis/build/expressjs/multer
> standard && mocha



  Disk Storage
    ✓ should process parser/form-data POST request
    1) should process empty fields and an empty file
    ✓ should process multiple files
    ✓ should remove uploaded files on error
    ✓ should report error when directory doesn't exist

  Error Handling
    ✓ should respect parts limit
    ✓ should respect file size limit
    ✓ should respect file count limit
    ✓ should respect file key limit
    ✓ should respect field key limit
    ✓ should respect field value limit
    ✓ should respect field count limit
    ✓ should respect fields given
    ✓ should report errors from storage engines
    ✓ should report errors from busboy constructor
    ✓ should report errors from busboy parsing

  Expected files
    ✓ should reject single unexpected file
    ✓ should reject array of multiple files
    ✓ should reject overflowing arrays
    ✓ should accept files with expected fieldname
    ✓ should reject files with unexpected fieldname

  Express Integration
    ✓ should work with express error handling (42ms)
    ✓ should work when receiving error from fileFilter

  Fields
    ✓ should process multiple fields
    ✓ should process empty fields
    ✓ should not process non-multipart POST request
    ✓ should not process non-multipart GET request
    ✓ should handle basic keys
    ✓ should handle multiple values
    ✓ should handle deeper structure
    ✓ should handle sparse arrays
    ✓ should handle even deeper
    ✓ should handle such deep
    ✓ should handle merge behaviour
    ✓ should handle bad fields
    ✓ should convert arrays into objects

  File Filter
    ✓ should skip some files
    ✓ should report errors from fileFilter

  File ordering
    ✓ should present files in same order as they came

  Functionality
    ✓ should upload the file to the `dest` dir
    ✓ should rename the uploaded file
    ✓ should ensure all req.files values (single-file per field) point to an array
    ✓ should ensure all req.files values (multi-files per field) point to an array
    ✓ should rename the destination directory to a different directory

  Issue #232
    ✓ should report limit errors

  Memory Storage
    ✓ should process multipart/form-data POST request
    ✓ should process empty fields and an empty file
    ✓ should process multiple files

  Reuse Middleware
    ✓ should accept multiple requests (48ms)

  Select Field
    ✓ should select the first file with fieldname
    ✓ should select all files with fieldname

  Unicode
[...]

PR #787, opened 2019-10-11
use 'close' event on file stream for reliability

Trott:
Change listened-for event on the output stream in storage/disk.js from
'finish' to 'close' to improve reliability. With 'finish', there is a
race condition such that sometimes the callback runs before the file
exists.

Refs: https://github.com/expressjs/multer/issues/238

### DIFF
--- a/storage/disk.js
+++ b/storage/disk.js
@@ -39,7 +39,7 @@ DiskStorage.prototype._handleFile = function _handleFile (req, file, cb) {
 
       file.stream.pipe(outStream)
       outStream.on('error', cb)
-      outStream.on('finish', function () {
+      outStream.on('close', function () {
         cb(null, {
           destination: destination,
           filename: filename,


Trott:
/ping @mcollina @ronag Does this seem like the correct event to listen for in this situation, or was the previous code correct and this is indicative of a more elusive problem?

ronag:
That's interesting. `'finish'` should be fine. I will dig into this. 

In the meantime this workaround is not entirely correct either, you might end up invoking the callback twice `'close'` will be emitted after `'error'`. Also you are not handling the `'error'` event from the source stream.

```js
      cb = once(cb)
      var finalPath = path.join(destination, filename)
      var outStream = fs.createWriteStream(finalPath)

      file.stream.pipe(outStream)
      file.stream.on('error', err => {
        outStream.destroy()
        cb(err)
      })
      outStream.on('error', cb)
      outStream.on('close', function () {
        cb(null, {
          destination: destination,
          filename: filename,
          path: finalPath,
          size: outStream.bytesWritten
        })
      })
```

ronag:
Yes, I think this might be a bug in node. Streams can emit `'finish'` before `'open'`, thus causing a race condition.

ronag:
https://github.com/nodejs/node/pull/29930

Trott:
> That's interesting. `'finish'` should be fine. I will dig into this. What version of node?

@ronag I was using Node.js 12.11.1.

Trott:
> [nodejs/node#29930](https://github.com/nodejs/node/pull/29930)

Awesome and thank you! I'll close this.

Issue #411, opened 2016-10-24
Empty `body` and `files` request properties


jamsinclair:
I feel like I must be doing something extremely silly here, but can replicate my problem in a really simple app following your examples. Perhaps there is something really weird with my node setup.

__Platform:__ OSX El Capitan 
__Node Version:__ v5.12.0 (I've also nvm-ed and tried with v6.7.0, same outcome)

__Expected Result:__
POST `multipart/form-data` requests with the multer middleware should add `body` and `files` properties to the request object with the data. 

__What happens for me:__
Empty `req.body` or `req.files` data from POST requests. Multer throws no errors and returns empty object and array for them respectively.

I've made a simple repo that shows my setup: https://github.com/jamsinclair/multer-test

My server js is literally:
(I've tried using different api methods, `.array()`, `.single('name')`, `.none()` etc, still the same outcome)
```javascript
var app = require('express')();
var multer = require('multer');

var upload = multer({ dest: './uploads' })

app.post('/upload', upload.any(), function (req, res, next) {
  console.log('body data:', req.body);
  console.log('files data:', req.files);
  res.sendStatus(200);
});

app.listen(9000, function () {
  console.log('Listening on port 9000');
});
```

Example payload being sent via Postman:
```
POST /upload HTTP/1.1
Host: localhost:9000
Content-Type: multipart/form-data; boundary=----WebKitFormBoundary7MA4YWxkTrZu0gW
Cache-Control: no-cache
Postman-Token: 40d74092-2b02-12c4-b213-6ecd515ff61f

------WebKitFormBoundary7MA4YWxkTrZu0gW
Content-Disposition: form-data; name="foo"

bar
------WebKitFormBoundary7MA4YWxkTrZu0gW
Content-Disposition: form-data; name="name"

multer
------WebKitFormBoundary7MA4YWxkTrZu0gW--
```

crowmagnumb:
I am having this exact same issue. node 6.7.0, multer 1.1.0, express 4.14.0, body-parser 1.15.2, cors 2.7.1 (just in case that affects things)

jamsinclair:
I can't quite figure it out, but have a feeling this is due to the way the request is being sent with Postman or Node, unrelated to `multer`.

A simple html form posts okay and outputs the body data correctly in the console 👌 
```html
<form action="./upload" method="POST" enctype="multipart/form-data">
  <input type="text" name="foo">
  <input type="submit">
</form>
```
Will update my example repo. 

[...]

PR #789, opened 2019-10-13
add execute request pipe. Fixes #572, #411

### LINKED ISSUE #572: empty req.body and req.file when run in firebase hosting
I don't know if its the problem of firebase or multer .
```nodejs

app.post('/' ,  multer({ dest: "data/" , limits : { fileSize : 5242880 } }).single('image_file'),  (req , res)=>{

    console.log("req.body : " , req.body) ;
    console.log("req.file  : " , req.file) ; 

```

It works fine on my local server but when I deploy it to firebase , it gives me empty req.body and req.file . How to fix this ?

EstebanFuentealba:
There was a breaking change in the Cloud Functions setup that triggered this issue. It has to do with the way the middleware works that gets applied to all Express apps (including the default app) used to serve HTTPS functions. Basically, Cloud Functions will parse the body of the request and decide what to do with it, leaving the raw contents of the body in a Buffer in req.rawBody. 

I added a parameter in the configuration to be able to manipulate the call to busboy. Fixes #572, #411

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ app.post('/profile', function (req, res) {
   })
 })
 ```
-## Custom request pipe busboy (Example firebase + multer + firebase storage)
+## Custom request pipe busboy (Example firebase + express + multer + firebase storage)
 ```javascript
 const admin = require("firebase-admin");
 const express = require("express");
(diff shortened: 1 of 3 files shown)


paaKways:
why hasn't this been merged already though?

reinaldoacdc:
This would be great.

UlisesGascon:
Superseded by #1466

PR #796, opened 2019-10-22
Add `startProcessing` config for pre-processed request objects

emadalam:
The following has been added to allow working with pre-processed request objects, eg those in cloud functions and similar environments.

-  ✨  `startProcessing` config
- 📚 Documentation update and sample usage
- 🚨 Tests to cover existing and updated use cases

Example:

```javascript
var app = express()
var upload = multer({
  startProcessing (req, busboy) {
    if (req.rawBody) { // indicates the request was pre-processed
      busboy.end(req.rawBody)
    } else {
      req.pipe(busboy)
    }
  }
})

app.post('/profile', upload.none(), function (req, res, next) {
  // req.body contains the text fields
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,27 @@ app.post('/profile', upload.none(), function (req, res, next) {
 })
 ```
 
+Custom `startProcessing` function to handle pre-processed request object, useful when dealing with cloud functions or environments that processes http `request` object before it can hit the express application.
+
+```javascript
+var express = require('express')
+var app = express()
+var multer  = require('multer')
+var upload = multer({
+  startProcessing (req, busboy) {
+    if (req.rawBody) { // indicates the request was pre-processed
+      busboy.end(req.rawBody)
+    } else {
+      req.pipe(busboy)
+    }
+  }
+})
+
+app.post('/profile', upload.none(), function (req, res, next) {
+  // req.body contains the text fields
+})
+```
+
 ## API
 
 ### File information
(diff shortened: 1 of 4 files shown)


TheAmazingPT:
Will this be merged?
I am using this fork with good success and `multer` really needs something to support Google Cloud Function requests :S

emadalam:
@TheAmazingPT Same here, I've been using this myself on production since the day I created this PR, no fallouts or errors so far.

I've added everything possible, tests, documentation etc, to make this PR ready for review and merge. It's up to the library maintainers to decide whether they want to review and merge this or provide something else to merge this 🤷‍♂

gmarcz:
It would be nice to have something (or similar) like this.
It could be a simple option useRawBody which is default false,
and a few line of code in the make-middleware.js

```
if (options.useRawBody && req.rawBody) {
  busboy.end(req.rawBody)
} else {
  req.pipe(busboy)
}
```

With such an option we could use multer with google cloud functions and don't have to reinvent the wheel.

ezzabuzaid:
Will this be merged anytime soon?

ezzabuzaid:
any updates on this...

LinusU:
Sorry for not seeing this earlier. I'm not sure that I'm comfortable adding this to Multer since it exposes a lot of the internals, which means that we won't be able to change it without it being a breaking change.

Is this really the only way that Express works on Google Cloud Function? Won't every single middleware need to be updated then? How does `express.json()` and similar middleware deal with this?

Another way that wouldn't expose any internals would be to add support for `rawBody` internally in the library. But I still think it's very strange that GCF took the decision to completely change how the body works. Is this documented somewhere?

ezzabuzaid:
Hi @LinusU

I had to build an abstraction over busboy directly to have it work. It's been a while for this issue so I do not remember why Multer didn't work before. You might refer to this page [GCF busboy](https://cloud.google.com/functions/docs/writing/http#multipart_data) _the last line_ which says more about why this PR is needed.

laruss:
any updates here?

UlisesGascon:
Superseded by #1466

PR #798, opened 2019-10-27
added conflict note to README

C0pyR1ght:
Had the issue of trying to use Multer in combination with express-fileupload. After hours of looking though issues and trying I got it working.
Thought this small hint might be helpful to safe others hours of debugging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,9 @@
 Multer is a node.js middleware for handling `multipart/form-data`, which is primarily used for uploading files. It is written
 on top of [busboy](https://github.com/mscdex/busboy) for maximum efficiency.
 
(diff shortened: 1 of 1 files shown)


helio-frota:
@C0pyR1ght hi,

While it is good to have some advice on this, and I believe your idea is to help other users, on the other hand I don't see it as something focused on the multer to the point that needs to add a note in README.

In fact multer functionality is being affected by the use of express-fileupload, as seen in this branch:
https://github.com/helio-frota/hold-my-file/blob/express-file-upload-conflict/server.js#L11

But let's imagine that other people write other module that deal with file uploads and that somehow can affect the multer if it is used together.

Another interesting example would be the use of a library that handles database connections, it would not make much sense to use two similar libraries to access the same database.
Something may have a good chance of going wrong.

PR #826, opened 2020-01-17
Use make-dir instead of mkdirp

### LINKED ISSUE #139: Invalid dest causes runaway node process
Given the following code:

```
var express = require("express"),
    multer = require("multer"),
    app = express();

app.use(multer({dest: "h:\\drive\\does\\not\\exist"}));
```

I would expect an exception to be thrown stating that the destination does not exist.

Instead, the node process goes to 100% CPU and starts a runaway memory leak that very quickly locks up the system if not killed.

jonchurch:
This PR is meant to fix #139

This PR replaces [`mkdirp`](https://www.npmjs.com/package/mkdirp) which doesn't seem to be maintained anymore with [`make-dir`](https://www.npmjs.com/package/make-dir).

I realize that v2 of multer likely [won't use storage engines at all](https://github.com/expressjs/multer/pull/399), but maybe this can make it into v1.x.

All tests pass, and I don't believe this should be breaking for anyone, as it should function exactly the same as `mkdirp` did previously. `make-dir` was created to replace `mkdirp`.

I also added a test to make sure that a path is created when `diskStorage` is given a string for `destination`, as per the docs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "append-field": "^1.0.0",
     "busboy": "^0.2.11",
     "concat-stream": "^1.5.2",
-    "mkdirp": "^0.5.1",
+    "make-dir": "^3.0.0",
     "object-assign": "^4.1.1",
     "on-finished": "^2.3.0",
     "type-is": "^1.6.4",
--- a/storage/disk.js
+++ b/storage/disk.js
@@ -2,7 +2,7 @@ var fs = require('fs')
 var os = require('os')
 var path = require('path')
 var crypto = require('crypto')
-var mkdirp = require('mkdirp')
+var makeDir = require('make-dir')
 
 function getFilename (req, file, cb) {
   crypto.pseudoRandomBytes(16, function (err, raw) {
@@ -18,7 +18,7 @@ function DiskStorage (opts) {
   this.getFilename = (opts.filename || getFilename)
 
   if (typeof opts.destination === 'string') {
-    mkdirp.sync(opts.destination)
+    makeDir.sync(opts.destination)
     this.getDestination = function ($0, $1, cb) { cb(null, opts.destination) }
   } else {
     this.getDestination = (opts.destination || getDestination)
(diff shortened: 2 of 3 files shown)


jonchurch:
Aww heck, I didn't realize this uses `async` under the hood, so it will only work on Node >=8 😞

LinusU:
Yeah, unfortunately this would be a breaking change, and thus we cannot add it to 1.x 😬 

If you any other library we could use I'd be happy to consider it, but we cannot break compatibility with Node.js 0.10.0.

jonchurch:
I [tweeted at](https://twitter.com/jonchurch/status/1218301090673889281) the maintainers of mkdirp to see if they'd consider a new release to fix this. Likely won't go anywhere, but figured I'd ask nicely.

wesleytodd:
@isaacs do you think you could take a look at this?  Looks like this is the result of https://github.com/substack/node-mkdirp/issues/70 not being resolved.  Is there something we could do on this?  Or is the correct move to use `make-dir` in `1.x` and then move to node core `fs.mkdir()` recursive in `2.x`?

isaacs:
I think using `make-dir` is a reasonable approach.  I've wanted to take some time to update mkdirp, but other things have taken precedence.  I may get to it soon, but then again, I would've said that a few months ago, and here we are.  I might also just scrap it and migrate to make-dir myself, because there are just other things too care about.  And I believe substack is busy creating peer to peer maps of his tropical volcano island paradise.

jonchurch:
For context, it looks like @isaacs [forked mkdirp](https://github.com/isaacs/node-mkdirp) and published a new major to npm. That version supports >= 8.x same as `make-dir`. With multer v2 coming out soon, I don't think the linked issue will be fixed.

jonchurch:
I opened a PR w/ the fix for the issue this PR was attempting to circumvent on the newly maintained branch for the 0.x line of `mkdirp`. https://github.com/isaacs/node-mkdirp/pull/15

Issue #829, opened 2020-01-18
CI pipeline is too old / not exist test run !


3imed-jaberi:
Should update the nodejs versions and add test for CI..

jonchurch:
@LinusU What do you think about updating Node versions in the CI? Sounds like a reasonable idea, adding 12 and maybe Current.

PR #832, opened 2020-01-18
Fix Test Cases [tests pass on all os]

### LINKED ISSUE #181: Test fails because of line endings on windows
Because most of our test files is ascii text files, they don't have the expected file size when checked out on window because git converts the line endings.

Simple fix is to replace them with random binary data, which makes sense because they already have `.dat` as their file-ending.

### LINKED ISSUE #827: Test Cases not work & & Coverage Code not exist !!
This test is not the only one that fails, if you fix it, it will result in failure in other tests .. 
As for Coverage, it's not exsit ..

image

3imed-jaberi:
- fix all test case .. 

fixed-test-case

- add the coverage script by add ~`nyc`~ `istanbul`.

istanbul-coverage

>
> `nyc`.. 
>
> add-coverage
>

- update the ci pipeline.
- cleanup .gitignore + locks.

merge #831 #830 #829 here .. 
fix #181 #827 #829 ..


3imed-jaberi:
> Thank you very much for contributing! First for feedback, this PR changes too many things to be easily evaluated.
> 
> Can you open an issue describing the things you are trying to fix or enhance individually? That makes it much easier to track discussions and focus on individual issues.
> 
> For example, I can't reproduce the tests failing. It might be a good idea to add more Node versions into CI, but that's it's own issue alone. As is adding code coverage. There's also formatting changes, this introduces a yarn.lock file, and ignores and disables package-lock files.
> 
> Also, there are 33 people who have contributed to this package in some way, but you'll notice that only 3 are listed as contributors in `package.json`. I'm not the person who decides who gets listed there, or how that works, but I wanted to point that out 😉

@jonchurch, thank you for your answer ..
I gathered them here because I see that they are related to each other .. If you want to separate them, tell me and I will do it right away ( you can re-open #828 #830 #831) ..
 As for the issues, I have opened two issues: #827 and #829, which are about the tests that fail and the coverage. I also spoke about the continuous integration pipeline which is very old and has no actual test / coverage ..

Regarding the subject of adding my name, I used to think that every contributor added his name .. I did not know that someone was decide putting people there ..
I will wait for your decision to separate each pr or review this request, then I will delete my name ...

3imed-jaberi:
I will close this PR due to lack of interaction ..
If anyone need this PR, you can leave a comment and I will reopen.

PR #841, opened 2020-02-16
Add Math.random() to storage filename example

drazisil:
I was encountering an issue where two files had the same name due to being handled in the same millisecond. While not a bug, adding extra uniqueness may save someone else time troubleshooting 'missing' files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,8 @@ var storage = multer.diskStorage({
     cb(null, '/tmp/my-uploads')
   },
   filename: function (req, file, cb) {
-    cb(null, file.fieldname + '-' + Date.now() + '-' + Math.random())
+    const uniqueHash = Date.now() + '-' + Math.random().split('.')[1]
+    cb(null, file.fieldname + '-' + uniqueHash)
   }
 })
 
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ var storage = multer.diskStorage({
     cb(null, '/tmp/my-uploads')
   },
   filename: function (req, file, cb) {
-    cb(null, file.fieldname + '-' + Date.now())
+    cb(null, file.fieldname + '-' + Date.now() + '-' + Math.random())
   }
 })
 
--- a/README.md
+++ b/README.md
@@ -175,8 +175,8 @@ var storage = multer.diskStorage({
     cb(null, '/tmp/my-uploads')
   },
   filename: function (req, file, cb) {
-    const uniqueHash = Date.now() + '-' + Math.random().split('.')[1]
-    cb(null, file.fieldname + '-' + uniqueHash)
+    const uniqueSuffix = Date.now() + '_' + (Math.random().toString().split('.')[1])
+    cb(null, file.fieldname + '_' + uniqueHash)
   }
 })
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -175,7 +175,7 @@ var storage = multer.diskStorage({
     cb(null, '/tmp/my-uploads')
   },
   filename: function (req, file, cb) {
-    cb(null, file.fieldname + '-' + Date.now())
+    cb(null, file.fieldname + '-' + Date.now() + '-' + Math.random())
```
LinusU: Hmm, this will produce filenames like: `foo-1581931472193-0.5359727197825674`. I'm not sure how that is being handled since essentially the file extension of the file is now `.5359727197825674` 🤔

drazisil:
In linux there's no problem, but you are correct  that Windows will
probably scream. I'll edit to remove the period.

README.md
```diff
@@ -175,7 +175,8 @@ var storage = multer.diskStorage({
     cb(null, '/tmp/my-uploads')
   },
   filename: function (req, file, cb) {
-    cb(null, file.fieldname + '-' + Date.now())
+    const uniqueHash = Date.now() + '-' + Math.random().split('.')[1]
```
LinusU: Can we call this something other than "hash" since it isn't a [hash](https://en.wikipedia.org/wiki/Hash_function)?

LinusU: `Math.random().split('.')[1]` this will throw an error:

> TypeError: Math.random(...).split is not a function

README.md
```diff
@@ -175,7 +175,8 @@ var storage = multer.diskStorage({
     cb(null, '/tmp/my-uploads')
   },
   filename: function (req, file, cb) {
-    cb(null, file.fieldname + '-' + Date.now() + '-' + Math.random())
+    const uniqueHash = Date.now() + '-' + Math.random().split('.')[1]
```
```suggestion
    const uniqueHash = Date.now() + '-' + Math.random().toString().split('.')[1]
```

drazisil: @jonchurch I already did that, you somehow reviewed an old copy?

jonchurch: Yep looks like it, ignore this hehe

README.md
```diff
@@ -175,8 +175,8 @@ var storage = multer.diskStorage({
     cb(null, '/tmp/my-uploads')
   },
   filename: function (req, file, cb) {
-    const uniqueHash = Date.now() + '-' + Math.random().split('.')[1]
-    cb(null, file.fieldname + '-' + uniqueHash)
+    const uniqueSuffix = Date.now() + '_' + (Math.random().toString().split('.')[1])
```
LinusU: What do you think about `Math.round(Math.random() * 1E9)` instead of `(Math.random().toString().split('.')[1])`, I feel that it's a bit easier to understand?

LinusU: Why change from `-` to `_`?

README.md
```diff
@@ -175,8 +175,8 @@ var storage = multer.diskStorage({
...
+    cb(null, file.fieldname + '_' + uniqueHash)
```
LinusU: > ReferenceError: No variable named "uniqueHash"

README.md
drazisil: If that works, I'm ok with it.

drazisil: I've run into issues where Javascript sees a dash and tries to do math. I figured it was safer, but i can change it back if you want.

README.md
drazisil: I corrected this in a prior commit

LinusU:
Great, thanks!

PR #844, opened 2020-02-23
Update make-middleware.js

hanani8:
Displaying the preservedPath


jonchurch:
This change passes the boolean value from the configuration object into the file object. Can you explain what issue this change fixes for you?

hanani8:
@jonchurch  I tried to, though unsuccessfully, preserve the full path of the file being uploaded. It doesn't actually solve my original problem.

PR #847, opened 2020-02-27
Reduce dependencies

disizali:
type-is , had 2 unnecessary dependencies and unneeded functions and operations at itself

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -177,7 +177,10 @@ function makeMiddleware (setup) {
 }
 
 function isMultipart(req) {
-  if ( !req || !req.headers['content-type'] || RegExp(/^multipart\/form-data;/).test(req.headers['content-type']) ) {
+  if ( 
+    !req ||
+    !req.headers['content-type'] ||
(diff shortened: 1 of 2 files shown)


jonchurch:
Thanks for contributing to the project! The way Express and it's family of middlewares works is that functionality is broken up across three different organizations on github:

* [expressjs](https://github.com/expressjs/express)
* [pillarjs](https://github.com/pillarjs)
* [jshttp](https://github.com/jshttp)

This is to make each piece modular and self contained. With the ability to be maintained and updated independently of Express. But most importantly, to share common functionality across the different projects so each one doesn't have to maintain it's own versions of core pieces. 

The module [type-is](https://github.com/jshttp/type-is) is one example. Removing it from this package means that any updates to how content type headers are handled in the future would need to be duplicated here in Multer.

It does do a very tiny piece of work, and can be replaced with a custom function like you have done, but that means moving Multer away from the shared way that all the pieces of express check content types. I hope I've explained why this particular module is actually useful ❤️ 

Closing this for now, but please do feel free to contribute in the future! This repo is one of the most active in the Express family and we are always looking for help.

PR #848, opened 2020-02-28
add stricter default limits

ryhinchey:
Issue https://github.com/expressjs/multer/issues/182#issuecomment-592418154

Adds stricter default limits for `fields` and `fileSize`:

Other library limits as reference:
`fields`: multiparty: 100, node-formidable: 1000
`fileSize`: multiparty: Infinity, node-formidable: 200mb

https://github.com/node-formidable/node-formidable/blob/master/src/Formidable.js#L14-L22
https://github.com/pillarjs/multiparty/blob/master/index.js#L61-L63

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,8 +139,8 @@ Key | Description | Default
 --- | --- | ---
 `fieldNameSize` | Max field name size | 100 bytes
 `fieldSize` | Max field value size | 1MB
-`fields` | Max number of non-file fields | Infinity
-`fileSize` | For multipart forms, the max file size (in bytes) | Infinity
+`fields` | Max number of non-file fields | 100 
+`fileSize` | For multipart forms, the max file size (in bytes) | 200MB
 `files` | For multipart forms, the max number of file fields | Infinity
 `parts` | For multipart forms, the max number of parts (fields + files) | Infinity
 `headerPairs` | For multipart forms, the max number of header key=>value pairs to parse | 2000
--- a/index.js
+++ b/index.js
@@ -12,7 +12,11 @@ function _middleware (limits, fields, fileStrategy) {
 
 class Multer {
   constructor (options) {
-    this.limits = options.limits
+    this.limits = {
+      fileSize: 200 * 1024 * 1024,
+      fields: 100,
+      ...options.limits
+    }
   }
 
   single (name) {
--- a/test/limits.js
+++ b/test/limits.js
@@ -1,5 +1,6 @@
 /* eslint-env mocha */
 
+const assert = require('assert')
 const assertRejects = require('assert-rejects')
 const FormData = require('form-data')
 
@@ -17,5 +18,13 @@ describe('limits', () => {
       util.submitForm(parser, form),
       (err) => err.code === 'LIMIT_FILE_SIZE' && err.field === 'file'
     )
+  })
+
+  it('should apply default limits', () => {
+    const parser = multer()
+    assert.deepStrictEqual(parser.limits, {
+      fileSize: 209715200,
+      fields: 100
+    })
   })
 })


LinusU:
This is a great start! I think that we should aim to not have any limits defaulting to `Infinity`.

I also think that `200MB` is a bit high 🤔 

I'll also look around a bit for more libraries and what their defaults have. Specifically other platforms I think would be interesting to look at, e.g. Ruby/Rails & PHP.

ryhinchey:
I took a look through Django as that's a framework I've used in the past and here's what they do:

- filesize: no limits. need to add your own validation (which isn't very hard to do)
- fields: 1000 https://docs.djangoproject.com/en/3.0/ref/settings/#data-upload-max-number-fields

They do have a default limit for uploading a file to memory. If an upload is more than 2.5mb, it'll get streamed to the filesystem
https://docs.djangoproject.com/en/3.0/ref/settings/#std:setting-FILE_UPLOAD_MAX_MEMORY_SIZE

README.md
```diff
@@ -139,8 +139,8 @@ Key | Description | Default
 --- | --- | ---
 `fieldNameSize` | Max field name size | 100 bytes
 `fieldSize` | Max field value size | 1MB
-`fields` | Max number of non-file fields | Infinity
-`fileSize` | For multipart forms, the max file size (in bytes) | Infinity
+`fields` | Max number of non-file fields | 100 
+`fileSize` | For multipart forms, the max file size (in bytes) | 200MB
```
```suggestion
`fileSize` | For multipart forms, the max file size (in bytes) | 209715200 (200MB)
```
Let's be specific here about the default value `209715200 (200mb)`

jonchurch:
I can't recall seeing issues about this, but I just noticed that the docs for the limits aren't very specific about the type of value accepted and what the actual default is. 

I'm thinking that because multer is often used by beginners, being very explicit about the file/field sizes being Numbers represented in bytes, might be a service to folks.

LinusU:
I think that using the [`bytes`](https://www.npmjs.com/package/bytes) module and documentation accordingly could be a good idea ☺️

ryhinchey:
That’s a great idea!

jonchurch:
What type of version bump do you think would be appropriate for that? It doesn't have to be breaking, we can keep the default behavior if passing a number, but add the ability to convert human readable strings using the byte module.

I ask this because I know this PR is one of the last things to do before getting the next major out, and I don't want to block on that at all.

ryhinchey:
> What type of version bump do you think would be appropriate for that? It doesn't have to be breaking, we can keep the default behavior if passing a number, but add the ability to convert human readable strings using the byte module.

100% agree.  If limits land in general, adding more human friendly ways to express them wouldn't have to be a major change.

README.md
ryhinchey: @jonchurch I was being consistent with how the other default values are rendered.  The table has a `Default` column. Do you think that should be changed?

ryhinchey: Oh nevermind - you're suggesting adding the number of bytes.   that wasn't done for `fieldSize` so I was trying to be consistent

jonchurch: Yeah that's what made me think about the human readable strings. When scanning this document I can see someone using `'10mb'` as the value based on the default value (even though it definitely says that the value should be in bytes)

LinusU:
Yeah, that doesn't have to be breaking, the only thing I want to get in now is sensible defaults ☺️ 

Unfortunately I got stuck at work yesterday, hope to get some time for this soon! Sorry for the delay...

In the meantime, any information on what any other systems are doing is much appreciated!

mercmobily:
I am the last kid in the block now, but can I add... please clarify in the docs how the fact that we now have the ability to have a *stream* changes the concept of "file size limit". Isn't the stream writer supposed to check that? How does it happen? Or are we basing the limit on header information? What if the header is wrong? This is probably all fully addressed by your capable hands -- I am just saying that it would be good to clarify it in the docs.

ryhinchey:
> In the meantime, any information on what any other systems are doing is much appreciated!

@LinusU I added the comment here about Django https://github.com/expressjs/multer/pull/848#issuecomment-592580375

Nginx has a `client_max_body_size` setting which defaults to 1mb: http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size

Looking through the docs for a popular form upload library for Flask, there are no defaults but if you call the default configuration, it defaults to 16mb.  https://pythonhosted.org/Flask-Uploads/

Looking through Go's http documentation and there are no default limits in place for `multipart/form-data` uploads

Looking through Spring Boot's docs, there's a file size limit of 1mb
https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/web/servlet/MultipartProperties.html

ryhinchey:
> I am the last kid in the block now, but can I add... please clarify in the docs how the fact that we now have the ability to have a _stream_ changes the concept of "file size limit". Isn't the stream writer supposed to check that? How does it happen? Or are we basing the limit on header information? What if the header is wrong? This is probably all fully addressed by your capable hands -- I am just saying that it would be good to clarify it in the docs.

this is a great point!  the limits are passed directly to `busboy` https://github.com/expressjs/multer/blob/explore-new-api/lib/read-body.js#L95-L127 and `multer` will throw an error if they're hit.  We can update the docs around this to point to the fact the limits are passed to busboy and outline how they're used there.

mercmobily:
Oh.. so busybox will actually count the bytes as they pass the stream?
So, if the uploaded file were to pass the limits, would this call
(responsible of writing the file's stream on the file system) actually fail?

      await pipeline(file.stream, fs.createWriteStream(`f/${record.id
}.${ext}`))

...? I think *this* is something that should go in the docs, for sure! (If
I am right, that is)

LinusU:
We should add a test to make sure that the error is definitely caught earlier, but the error _should_ be caught before since the file is temporarily stored in intermittent files in order to process the body stream before handing over from the middleware

LinusU:
Also, sorry for not getting time to close this one yet, hopefully I'll manage today! 👏

LinusU:
I'm going to just post as I find things, and then try and clean it up later 😁 

PHP:
- Max files: 20
- Max file size: 2M
- Max fields: 1000
- Entire post body: 8M
- Source: https://www.php.net/manual/en/ini.list.php

LinusU:
I cannot find any limits or ways to limit in Ruby on Rails, unless your adding a third party dependency to validate the files after 🤔 

Might be missing something here though...

LinusU:
| Name | Field Name | Field Size | Field Count | File Size | File Count | Part Count | Header Count | Entire Body |
| :---- | ----: | ----: | ----: | ----: | ----: | ----: | ----: | ----: |
| Multer 1.x | 100b | 1M | - | - | - | - | 2000 | - |
| PHP | - | - | 1000 | 2M | 20 | - | - | 8M |
| Nginx | - | - | - | - | - | - | - | 1M |

LinusU:
> They do have a default limit for uploading a file to memory. If an upload is more than 2.5mb, it'll get streamed to the filesystem

Hmm, interesting, but it seems like they also default to limiting the entire body size to 2.5mb? https://docs.djangoproject.com/en/3.0/ref/settings/#data-upload-max-memory-size

mercmobily:
> We should add a test to make sure that the error is definitely caught earlier, but the error _should_ be caught before since the file is temporarily stored in intermittent files in order to process the body stream before handing over from the middleware

Wait... so it's a stream, but the data is not "streaming" (as such) from the flow as data arrives, but from intermediate temporary files stored on the file system...?

LinusU:
> Wait... so it's a stream, but the data is not "streaming" (as such) from the flow as data arrives, but from intermediate temporary files stored on the file system...?

That's correct, the multipart/form-data format doesn't really allow for streaming parsing as information about each file comes after the each other, and interleaved with other form data.

LinusU:
Added quite a few things related to limits here: https://github.com/expressjs/multer/pull/399#issuecomment-599232466

Would love some feedback!

ryhinchey:
> Would love some feedback!

Looks good to me @LinusU!  I like that you removed `parts`.  I think it cleans up the limits interface nicely :)

ryhinchey:
@LinusU I’m going to close this PR in favor of the work you did in https://github.com/expressjs/multer/pull/399

PR #875, opened 2020-04-22
Support Field Filter

noonii:
- Equivalent to file filter, except for fields

Because multer parses FormData based on the order of inputted values, sometimes file is parsed before fields. Ideally, if we can filter fields and reject before file upload that would be best.
```
const formData = new FormData();
formData.append('file', file, file.name);
formData.append('field', 'value');
```

Even so, we want to filter/validate fields and abort if necessary.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,11 @@ var diskStorage = require('./storage/disk')
 var memoryStorage = require('./storage/memory')
 var MulterError = require('./lib/multer-error')
 
-function allowAll (req, file, cb) {
+function allowAllFiles (req, file, cb) {
+  cb(null, true)
+}
+
+function allowAllFields (req, fieldname, value, cb) {
   cb(null, true)
 }
 
@@ -19,7 +23,8 @@ function Multer (options) {
 
   this.limits = options.limits
   this.preservePath = options.preservePath
-  this.fileFilter = options.fileFilter || allowAll
+  this.fileFilter = options.fileFilter || allowAllFiles
+  this.fieldFilter = options.fieldFilter || allowAllFields
 }
 
 Multer.prototype._makeMiddleware = function (fields, fileStrategy) {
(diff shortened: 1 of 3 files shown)


noonii:
@LinusU

noonii:
@irfanfs you pinged the closed PR I made (accident)

This is the right PR. If they merge this I imagine they'd pub 1.4.3 too.

irfanfs:
> @irfanfs you pinged the closed PR I made (accident)
> 
> This is the right PR. If they merge this I imagine they'd pub 1.4.3 too.

@noonii i want the bumped version of **"mkdirp"** to fix snyk vulnerability issue

LinusU:
@irfanfs 

> i want the bumped version of "mkdirp" to fix snyk vulnerability issue

run `npm update mkdirp --depth=10` to bump your version of `mkdirp`

-----

@noonii Would you mind expanding on the use case for this? None of the other express body parsers offers functionality to filter what goes into `req.body` as far as I know 🤔 

Maybe this is better handled by a form validation library?

noonii:
> @noonii Would you mind expanding on the use case for this? None of the other express body parsers offers functionality to filter what goes into `req.body` as far as I know 🤔
> 
> Maybe this is better handled by a form validation library?

Server side validation.

Use case no. 1
1. User populates formData with fields before files
2. Multer parses fields before files
3. fields parsed and throws validation error from filter
4. operation aborted and data removed

Use case no. 2
1. User populates formData with files before fields
2. Multer parses files before fields
3. files already stored in dir or stream
4. fields parsed and throws validation error from filter
5. operation aborted and data removed

Without the filter in all cases the file is uploaded to host and then leaving the app to validate and remove the file accordingly. The code to handle all of this efficiently exists in multer already, so it becomes redundant to the app. I suggest removing the redundancy. 

Bonus: only in use case no.1 would it serve best because it handles aborting before the upload.

I run a service at scale that would benefit from this.

panzi:
If this field filter would also get the mime type of the field and could transform the value (i.e. the callback should get a 3rd parameter: the transformed value) than my pull request would be obsolete: https://github.com/expressjs/multer/pull/1037

PR #907, opened 2020-06-07
Add preserveBody option to avoid overwriting request body

### LINKED ISSUE #445: Custom middleware request body gets overriden
I have custom middleware that appends request body and puts some value there. But when is form-data, that appended value dissapears from request body. I did a bit of investigation around and i found why.

In multer file `lib/make-middleware.js` at line 27, there is this piece of code `req.body = Object.create(null)` which destroys previous set values by my middleware. I simply commented this and solved problem for me. I didn't go further to investigate why this piece of code exists and purpose of it.

george3447:
Fixes #445 
cc: @LinusU @jonchurch

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ Key | Description
 `fileFilter` | Function to control which files are accepted
 `limits` | Limits of the uploaded data
 `preservePath` | Keep the full path of files instead of just the base name
+`preserveBody` | Extend existing request body, by default it is false
 
 In an average web app, only `dest` might be required, and configured as shown in
 the following example.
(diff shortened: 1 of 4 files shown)


maneetgoyal:
Would be nice to have this feature 😺

iamdenish (approved):

SeriousMonk:
2 years passed and this is still not merged. Why?

eboody:
any news on this?

vandermnt:
@puthawaladenish because it hasn't been merged yet?

UlisesGascon:
Thanks @george3447, and sorry this sat so long. We won't be merging this one: `req.body` is the output of multer's own parsing (like every body-parser), so extending an existing body means client-sent fields merge into values set by earlier middleware (`append-field` turns a collision into an array), and dropping the null-prototype body reopens the `__proto__` field-name problem we fixed. Data that doesn't come from the request body is better placed on another property (e.g. `req.user`) or added by a middleware that runs after multer.

Closing #445 with the same reasoning.

PR #971, opened 2021-01-14
Close write stream on connection error (FD leak on incomplete uploads)

Khez:
We have seen a large number of open FD's when using express with multer.

After investing, we have found the issue to be aborted uploads.

It seems the upload FD is not being closed if the client aborts the request mid-way.

## How To Reproduce
From the [readme](https://github.com/expressjs/multer#usage) (adapted to a single field):

```javascript
const express = require('express');
const multer  = require('multer');
const upload = multer({ dest: 'uploads/' });

var app = express();
var cpUpload = upload.fields([{
    name: 'avatar',
    maxCount: 1
}]);

app.post('/upload', cpUpload, function (req, res, next) {
    res.send('ok');
});

app.listen(8081, () => {
    console.log('Listening');
});
```

Using whichever client, upload a file and abort it during transit:

```bash
# make sure the file is somewhat large
# and don't forget to abort before getting the response
curl -vvv 'http://localhost:8081/upload' -F 'avatar=@file' --limit-rate 1M;
```

Finally using [lsof](https://en.wikipedia.org/wiki/Lsof) on the node process, you should see a FD leaked to the `uploads` directory

## Related(?)
* https://github.com/nodejs/node/issues/7732
    * Really old bug, closed without being solved, seems like this is is the "core" issue
* https://github.com/fastify/fastify-static/issues/116#issuecomment-655529420
    * Similar reproduction steps, but on downloading files

### DIFF
--- a/storage/disk.js
+++ b/storage/disk.js
@@ -40,9 +40,11 @@ DiskStorage.prototype._handleFile = function _handleFile (req, file, cb) {
 
       file.stream.pipe(outStream)
 
-      connection.on('error', function(error) {
-        outStream.end();
-      })
+      if (connection) {
+        connection.on('error', function(error) {
+          outStream.end();
+        })
+      }
 
       outStream.on('error', cb)
       outStream.on('finish', function () {
--- a/storage/disk.js
+++ b/storage/disk.js
@@ -36,8 +36,14 @@ DiskStorage.prototype._handleFile = function _handleFile (req, file, cb) {
 
       var finalPath = path.join(destination, filename)
       var outStream = fs.createWriteStream(finalPath)
+      var connection = req.socket || req.connection
 
       file.stream.pipe(outStream)
+
+      connection.on('error', (error) => {
+        outStream.end();
+      })
+
       outStream.on('error', cb)
       outStream.on('finish', function () {
         cb(null, {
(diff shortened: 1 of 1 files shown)


Khez:
I'm saddened to see this ignored. This was a very visible issue and with a clear solution.

We attempted a similar fix in production, but we still had some FD Leaks, uncertain why. It looked something like:

```js
const express = require('express');
const multer  = require('multer');
const upload = multer({ dest: 'uploads/' });

var app = express();
var cpUpload = upload.fields([{
    name: 'avatar',
    maxCount: 1
}]);


app.post('/upload', function (req, res, next) {
    // Userland fix for https://github.com/expressjs/multer/pull/971
    (req.connection || req.socket).on('error', (error) => {
        console.log('Connection Error Detected', error.message || error);
        req.emit('end');
    });

    cpUpload(req, res, function(err) {
        console.log(err || 'Upload Done');
        res.send('ok');
    });
});

app.listen(8081, () => {
    console.log('Listening');
});
```

Due to this PR was gathering dust and the fact that we still had FD leaks, we ended up moving to [formidable](https://github.com/node-formidable/formidable). Haven't had any leaks since then.

binSaed (approved):

UlisesGascon:
Thanks @Khez! This was fixed in multer 2.3.0 as part of https://github.com/expressjs/multer/security/advisories/GHSA-qfvm-cv95-jqjf, so I'm closing this one.

Issue #883, opened 2020-04-30
[docs, help wanted] Frontend Usage Examples


jonchurch:
We often see Issues opened for folks who are having troubles on the client side. Uploading an array of files clientside is non-obvious, due to no fault of Multer.

Most recently #880, but there are many such examples. (@expressjs/triagers maybe link some related issues here if you find them?)

PR https://github.com/expressjs/multer/pull/580 addressed this with a regular HTML form. 

**I propose a specific section of the docs, something like `## Clientside Examples` to feature prominently, and allow easy linking to, with common examples.**

Specifically we are missing an example for [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) web api. We can't cover everything, but let's cover some of the most asked cases.

mvasigh:
I'm happy to take this on. I can add examples for single and multiple files using HTML form and the `FormData` API, and perhaps some CodeSandbox links to examples with some popular front-end frameworks?

jonchurch:
Great to hear @mvasigh! Let's tackle one example at a time, it makes reviewing PRs much easier. We're currently missing a `FormData` example, so why don't you start there? We have an html form example in the docs right now, but we can expand on that soon. 

Let's also try and keep examples [Minimal](https://stackoverflow.com/help/minimal-reproducible-example).

@LinusU What do you think about where these examples should live? Maybe an examples directory?

lucasraziel:
It seems this issue is still open, I will try to add a React example

PR #998, opened 2021-04-12
Update first example of frontend usage. 

lucasraziel:
This is a first example to help in #883  issue. But I do not know if this one example would be enough to close issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,6 +322,68 @@ app.post('/profile', function (req, res) {
 })
 ```
 
+### Client side examples
+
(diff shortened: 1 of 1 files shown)


UlisesGascon:
Thanks @lucasraziel! Client-side examples are landing as runnable projects in the `examples/` folder (#896) rather than in the README, which stays framework-neutral. If you'd like to contribute a React variant there once that's merged, it would be welcome as its own directory. Closing this one.

PR #1006, opened 2021-05-06
Fix in multer configuration (Storage)

psimanta:
I tried with settings mentioned in the README.md file.
If I use the slash like this /tmp it refers my c drive, but if I don't use slash it works.


LinusU:
The intention of the example was to put the file in the root temp folder, so I'm not sure that this is the best approach 🤔 

Maybe switch it to use something like: `path.join(os.tmpdir(), 'my-uploads')`?

PR #1024, opened 2021-08-06
doc: update var to const in all Readmes

### LINKED ISSUE #1012: [docs] [question] Why does the readme use var in place of let / const ?
Multer is a **very** popular library, and the readme is even imported in other reputed projects (like express.js docs, link:http://expressjs.com/en/resources/middleware/multer.html).

Is there any particular reason we're using var in the readme? Almost all editors and linters would beserk out on this.

Should we change it to const/let as applicable?

jonchurch:
Update `var` to `const` in Readme and all translations. This was done with a search replace on each document, I believe but did not verify that all code examples are the same in all translations. 

Interestingly though, I think the English readme might have an extra codeblock that the others don't. Someone should check the blame and see if we need to update the translations with an example. Although I guess we would then need to translate that example haha. 

Edit: [Found it](https://github.com/expressjs/multer/pull/580) and it's of course something I reviewed. Looking back at it, I'm not sure that example was necessary. I think it just repeats the example above it. Maybe the elaborated explanation should get moved up.

closes #1012

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,11 @@ Don't forget the `enctype="multipart/form-data"` in your form.
 ```
 
 ```javascript
-var express = require('express')
-var multer  = require('multer')
-var upload = multer({ dest: 'uploads/' })
+const express = require('express')
+const multer  = require('multer')
+const upload = multer({ dest: 'uploads/' })
 
-var app = express()
+const app = express()
 
 app.post('/profile', upload.single('avatar'), function (req, res, next) {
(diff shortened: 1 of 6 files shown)


LinusU (approved):
Awesome 👏

LinusU:
> Looking back at it, I'm not sure that example was necessary. I think it just repeats the example above it. Maybe the elaborated explanation should get moved up.

Yeah, I think you are right, I'm open to changing it 👍

PR #1037, opened 2021-10-01
add parseJsonFields option

panzi:
Fields may also have a `Content-Type` header. If you set `parseJsonFields` to
`true` these fields will be parsed using `JSON.parse()` instead of handled as
plain text strings. This way you don't need to unroll complex JSON structures
that are transmitted alongside uploaded files as url-encoded fields.

If this is not good for some reason I'd at least like to have some way to get
what the content-type of the field was in the request handler, so I can at least
handle that situation manually.

A possible extension: Make it `parseFields` instead and not just handle
`application/json` content-types, but also fields with a
`application/x-www-form-urlencoded` content-type, or instead of a boolean
flag make it possible to pass a filter function (`value = fieldFilter(value, mimetype)`).
An advantage to the filter function approach would be the possibility to throw
a custom error type which can be handled more specifically than a `SyntaxError`
in the servers error handler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ Key | Description
 `dest` or `storage` | Where to store the files
 `fileFilter` | Function to control which files are accepted
 `limits` | Limits of the uploaded data
+`parseJsonFields` | Parse fields that have the content-type `application/json`
 `preservePath` | Keep the full path of files instead of just the base name
 
 In an average web app, only `dest` might be required, and configured as shown in
@@ -303,6 +304,13 @@ function fileFilter (req, file, cb) {
 
 }
 ```
+
+### `parseJsonFields`
+
+Fields may also have a `Content-Type` header. If you set `parseJsonFields` to
+`true` these fields will be parsed using `JSON.parse()` instead of handled as
+plain text strings. This way you don't need to unroll complex JSON structures
+that are transmitted alongside uploaded files as url-encoded fields.
 
 ## Error handling
 
--- a/index.js
+++ b/index.js
@@ -20,6 +20,7 @@ function Multer (options) {
   this.limits = options.limits
   this.preservePath = options.preservePath
   this.fileFilter = options.fileFilter || allowAll
+  this.parseJsonFields = !!options.parseJsonFields
 }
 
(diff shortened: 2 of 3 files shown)


noureldin-khaled:
Hi @panzi :wave: any ideas of if/when will this be released? I'm currently facing this problem and stuck with no solution other than trying unroll the JSON structure I have which will be a nightmare.
Ideally like you said there would be a generic `parseFields` but I would settle for just the `parseJsonFields` now

panzi:
@noureldin-khaled I don't know more than is documented here. I had this problem, so I forked multer to add this feature and made a pull request. Never heard anything back.

panzi:
If a field is sent as `Content-Type: application/json` I want to decode it correctly on the server side (versus when it is `text/plain` or has no header, where I handle it as a single string value). However, multer just drops the headers of fields and there is no way to handle that manually. Therefore I would like a feature where I could either supply a function to decode fields where the decoding function gets the mime type or a feature where mutler just decodes such fields automatically. Note that I'm talking about otherwise normal fields, not uploaded files. Fields that are supplied in addition to uploaded files.

LinusU:
Which client is sending you this data? Is this something that can be done from the browser?

It seems like this is almost like a request in the request, and potentially should be able to use any regular express body parser?

noureldin-khaled:
@LinusU Kindly consider including this PR with the next release, since multer overrides body parser when the request is multipart I feel like it should support all scenarios, including the scenario when a json object is sent in the multipart request, it should correctly parse it.

LinusU:
> [...] since multer overrides body parser when the request is multipart I feel like it should support all scenarios, including the scenario when a json object is sent in the multipart request, it should correctly parse it.

(first of all, sorry for the late response)

I still feel that this is outside of the scope of Multer.

With the same logic, one could say that Multer should also unzip files, parse xml, read image files into `ImageData`, etc.

What's the problem with just doing `const foo = JSON.parse(req.body.foo)` in your request handler?

panzi:
The problem is that I cannot access the `Content-Type` of the part at that point and thus don't know that it is `application/json`.

Issue #1041, opened 2021-10-27
Multer hangs


sivandu:
It seems a known issue with busboy module that was fixed:
https://github.com/mscdex/busboy/pull/149 
https://github.com/mscdex/busboy/issues/188

Can you please release a new version of Multer with the latest busboy version?

Issue #1042, opened 2021-11-02
Plans for v2?


recaph:
Hi Team, 

Our scans have identified a few dependencies as operational risk due to being old/unmaintained. These are coming from busboy and concat-stream dependencies.

multer v2.0.0rc3 seems to be addressing these already (busboy 0.3.1 and removing concat-stream)

![image](https://user-images.githubusercontent.com/704289/139947293-18b4d40b-37b3-4411-aa26-10b2920a8786.png)

It would be great to know when v2 will finally be released.

LinusU:
Right now I cannot commit to a specific date, but if you could test the latest release candidate and share your experience in #399 that would really help move this forward. As it is now I'm a bit afraid to release it since I haven't heard too much from people using the the v2 release candidates in production environments 😅

recaph:
Actually we depend on a different project https://github.com/nestjs/nest/tree/master/packages/platform-express which depends on older version of this project.. I can ask them too but they may not agree to use a rc release..

TeemuKoivisto:
@LinusU hey Linus, I just started debugging things and happened to stumble upon upgrading multer to `"^2.0.0-rc.3"`. Not having both commonjs and ESM imports is really a no-go for many projects. I can't just upgrade everything to use ESM and it's awkward to require an ESM import. I'd suggest using a bundler and adding both, "main" and "module" blocks. With Rollup for example that could be:

```
import commonjs from '@rollup/plugin-commonjs'

import pkg from './package.json'

export default {
  input: 'src/index.js',
  output: [
    {
      file: pkg.main,
      format: 'cjs',
    },
    {
      file: pkg.module,
      format: 'es',
    },
  ],
  external: [...Object.keys(pkg.dependencies || {})],
  plugins: [commonjs()],
}
```

Breaking things downstream is just really annoying and I much prefer not having to debug unrelated things to my current task.

LinusU:
> I'd suggest using a bundler and adding both, "main" and "module" blocks.

The problem is that our dependencies are ESM-only, so the package cannot be CJS and import those modules (unless we somehow require them asynchronously, which could possible work 🤔. Although at this point it seems like the entire ecosystem will move to ESM sooner or later so 🤷 )

> Breaking things downstream is just really annoying and I much prefer not having to debug unrelated things to my current task.

The entire API has changed anyways, since we have removed the concept of storage engines and are working with streams instead, so it will be a breaking change even without ESM...

TeemuKoivisto:
[...]

PR #1044, opened 2021-11-08
chore: update busboy

Achneoder:
Updates busboy to latest version 

Targets `DeprecationWarning: Buffer()` and #1041

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   ],
   "dependencies": {
     "append-field": "^1.0.0",
-    "busboy": "^0.2.11",
+    "busboy": "^0.3.1",
     "concat-stream": "^1.5.2",
     "mkdirp": "^0.5.4",
     "object-assign": "^4.1.1",
@@ -39,7 +39,7 @@
     "testdata-w3c-json-form": "^1.0.0"
   },
   "engines": {
-    "node": ">= 0.10.0"
+    "node": ">= 4.5.0"
   },
   "files": [
     "LICENSE",


LinusU:
Since this is a breaking change we cannot release it in a 1.x version, and I believe that this have already been updated in the 2.x line (see #399).

Thanks for taking the time with the PR though ☺️

Achneoder:
I know that it's fixed in version 2, but since this is a potential security issue, I think it's worth to consider releasing a new version fixing this.

You mentioned yourself that you don't know when v2.0 gets released (https://github.com/expressjs/multer/issues/1042#issuecomment-961231821) and according to the commits, it looks like v2.0 is in development for ~ 5 years now.

Dropping support for really old and not yet anymore maintained Node versions by creation a new minor version for multer doesn't look like a big deal to me in contrast to the potential (and also annoying) security issue by using `new Buffer()` - which is also already deprecated since Node 6.
But to be honest, I don't have any usage statistics for different Node versions and it might be that Node versions between 0.10.0 and 4.5.0 are still heavily used.

LinusU:
> [...] the potential (and also annoying) security issue by using new Buffer() - which is also already deprecated since Node 6.

Using `new Buffer(...)` isn't inherently unsafe, it's just when it's being used in some specific ways. If you believe that there is actually a security issue here, please report it in accordance with this guide:

https://github.com/expressjs/express/blob/master/Security.md

> Dropping support for really old and not yet anymore maintained Node versions by creation a new minor version for multer doesn't look like a big deal to me [...]

I think that following semver is really important, especially for projects being used by so many. It's not fun to have your project break because a dependency wasn't following it...

> You mentioned yourself that you don't know when v2.0 gets released (#1042 (comment)) and according to the commits, it looks like v2.0 is in development for ~ 5 years now.

All the features for Multer 2.0 is done, and the release candidates should be stable. I can recommend using 2.x now, and report any feedback, good or bad, in that thread. Once I see that it's working for some people I would feel confident releasing it...

Issue #1046, opened 2021-11-11
next() function not called in middleware, which breaks express-http-context (cls-hooked)


kr99:
https://github.com/expressjs/multer/blob/63f487421baec6f20d6a0a38561698f80daa2cb6/lib/make-middleware.js#L176

The last line in the function **should** be a call to `next()` (right after the line shown)

When I do roughly something like this...
```
const httpContext = require('express-http-context')
app.use(httpContext.middleware)
app.use(() => {httpContext.set('thing', 'value')})
...
router.put('/', upload.single('file'), myFunction)
...
inside myFunction():
httpContext.get('thing') // undefined, but only when there's a 'file' param in the multipart form...
```

dougwilson:
Of course if there was a next() call at that location, then your myFunction would never have access to any uploaded files, as busboy would not have the opportunity to recieve the request data and parse the form.

kr99:
That's fair--I don't understand what exactly is happening here I guess.  Any pointers or suggestions @dougwilson ?  Thanks for your time.

dougwilson:
I'm not familiar with that context module and what it is doing to keep the state around. Have you opened an issue in their repo about the state being lost? It seems to be a pretty common issue opened there and I didn't see any regarding multer. If you or that module's author can provide some kind of PR or specific guidence on what to change in multer that would be a big help. It is also possible that it is just a change in that module, but that module's author would know better.

PR #1047, opened 2021-11-11
fix bug where next() is not called, which breaks cls-hooked (express-http-context)

kr99:
See https://github.com/expressjs/multer/issues/1046


jetpack-toast (approved):

cuporter (approved):

kr99:
This actually breaks multer. Sorry for the bad PR.

PR #1067, opened 2022-01-25
DiskStorage destination folder check for when the destination is passed as a function.

AbhilakshSinghReen:
The 'DiskStorage' function in 'storage\disk.js' will now check if the destination folder exists on the disk or not. Earlier this check was only being performed when the destination was passed as a string, not when the destination was passed as a function. Changes have also been made to throw a type error if the destination is a type other than a function or a string.

### DIFF
--- a/storage/disk.js
+++ b/storage/disk.js
@@ -14,14 +14,25 @@ function getDestination (req, file, cb) {
   cb(null, os.tmpdir())
 }
 
-function DiskStorage (opts) {
-  this.getFilename = (opts.filename || getFilename)
+function DiskStorage(opts) {
+  this.getFilename = opts.filename || getFilename;
+  const destinationDatatype = typeof opts.destination;
 
-  if (typeof opts.destination === 'string') {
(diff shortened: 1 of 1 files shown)


mriedem:
The README should be updated:

https://github.com/expressjs/multer/blob/master/README.md#diskstorage

>Note: You are responsible for creating the directory when providing destination as a function.

LinusU:
I'm not sure if we would be able to release this in 1.x since I think it's a breaking change?

And in 2.x we have removed `DiskStorage` 😅

AbhilakshSinghReen:
> I'm not sure if we would be able to release this in 1.x since I think it's a breaking change?
> 
> And in 2.x we have removed `DiskStorage` 😅

AbhilakshSinghReen:
Okay. No problem. Thanks for the update.

> I'm not sure if we would be able to release this in 1.x since I think it's a breaking change?
> 
> And in 2.x we have removed `DiskStorage` 😅

PR #1079, opened 2022-03-23
update mkdirp (deprecated version)

guicompeng:
mkdirp 0.5.4 is deprecated

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "append-field": "^1.0.0",
     "busboy": "^0.2.11",
     "concat-stream": "^1.5.2",
-    "mkdirp": "^0.5.4",
+    "mkdirp": "^0.5.6",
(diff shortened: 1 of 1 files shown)


LinusU:
Currently, we have an unfortunate situation where Multer 1.x has had a special `1.4.5-lts.1` release whilst waiting for someone to backport the fixes in Busboy 1.x. Because of that I would prefer to release as few patches as possible at the moment.

Since the new range is already covered by the old one, I don't think that there is any reason to bump this?

For anyone that is coming here to resolve the vulnerability warning, you just need to update your lock file to use version `0.5.6` of `mkdirp`. Run `npm audit fix`, or `npm update mkdirp --depth=16`, or simply delete your lock file and install your dependencies again.

I don't see that there is anything we need to to in Multer, but please correct me if I'm wrong. Closing this for now...

PR #1086, opened 2022-04-29
Added support for JSON fields for parity with Axios;

DigitalBrainJS:
Since `v0.27.2` Axios [supports](https://github.com/axios/axios#-automatic-serialization) FormData serializer that encodes object fields with `{}` ending to a JSON string. So this PR adds the ability to parse such fields by multer.
```js
await axios.postForm('http://localhost/', {
  str: 'Jay"
  'obj{}': {x:1, foo: 'bar'},
  'files[]': document.querySelector('#myFile').files
});
```
Will be stored as `req.body.obj`

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -9,6 +9,14 @@ var MulterError = require('./multer-error')
 var FileAppender = require('./file-appender')
 var removeUploadedFiles = require('./remove-uploaded-files')
 
+var endsWith = (function (endsWith) {
+  return endsWith ? function (str, ending) {
+    return endsWith.call(str, ending)
+  } : function (str, ending) {
+    return str.slice(-ending.length) === ending
+  }
+})(String.prototype.endsWith)
+
 function drainStream (stream) {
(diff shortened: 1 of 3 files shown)


LinusU:
I feel that this Axios-specific quirk is out of scope for Multer. Is this implemented by any other multipart parser or serialiser?

I think that this could be easily implemented as it's own middleware? Something like this:

```js
// Multer
app.use(upload)

// Custom middleware
app.use((req, res, next) => {
  for (const key of Object.keys(req.body)) {
    if (key.endsWith('{}')) {
      req.body[key.slice(0, -2)] = JSON.parse(req.body[key])
      delete req.body[key]
    }
  }
})
```

That code could also be published as a package, that would work with any form-data body parser!

LinusU:
I'm sorry, but I still feel that this is out of scope for Multer. If this is something that more clients than Axios implements (especially if `fetch` implements it) I'd be happy to reconsider.

Thanks for taking the time to submit the PR though 🙏

PR #1087, opened 2022-05-06
Bump mkdirp lib to resolve minimist vulnerability.

MdAbdulSamadBaig:


LinusU:
We cannot update to 1.x without a breaking change, since it doesn't support the same Node.js ranges as Multer does. Is this an issue that is not solved in mkdirp 0.5.6?

bottjert:
With the above comments in mind, this PR might be a better short-term fix for the vulnerabilities
https://github.com/expressjs/multer/pull/1079
Who/what are we waiting on for this to be actioned?

Issue #790, opened 2019-10-15
expose filename*


bejito:
In the Content-Disposition header spec, there is a field called `filename*` that allows to stransmit utf8 characters. 
Neither `fieldname` not `originalname` expose that field.
Is it possible to get the value in that field from the multer execution ?

bejito:
Actually, `originalname` seems to read it, but the value it exposes just contains bunch of � for non ascii charactes

jonchurch:
Hmmm, this interesting. 

Can you send a minimal reproduction of this issue? Multer uses [Busboy](https://github.com/mscdex/busboy) to deal with handling file uploads [under the hood](https://github.com/expressjs/multer/blob/805170c61530e1f1cafd818c9b63d16a9dd46c36/lib/make-middleware.js#L96). 

I'm not entirely sure if the `originalname` field is messed with anywhere along the way that could reproduce this issue. But if you can send an example maybe we can step through it and find out what's happening.

Thanks!

ryhinchey:
This might have something to do with file encoding.  See https://github.com/expressjs/multer/issues/434

https://github.com/expressjs/multer/pull/399 might solve this

erano067:
After some debugging, the problem is in Busboy:

![image](https://user-images.githubusercontent.com/49909926/168907385-4592c74c-3e35-43f9-822d-40ddcae91d16.png)

In my example the header of `'content-disposition'` is set to:
`form-data; name="doc"; filename="×××¤×¡101.pdf"; filename*=UTF-8''%D7%98%D7%95%D7%A4%D7%A1101.pdf`
making a double `filename` fields.

Then in the for loop it takes the ugly first `filename` and convert it to utf-8, in that stage the `filename` variable looks good.
**but then** it makes another iteration and overwrite `filename` with the result of this function:
`filename = decodeText(parsed[i][1], 'binary', 'utf8');`

when we call that function in the second time we send it a 'utf-8' string, and tells it to decode the text from source = 'binary' to target = 'utf-8'.
And that makes the ugly  � �question � � marks � �.

I will open a PR in busboy and then we need to bump multer's busboy dependency.

for now as a workaround I managed to send the filename alongside the file itself like this:

![image](https://user-images.githubusercontent.com/49909926/168909364-4cf15a01-3b6d-46e2-bb16-eb0dbfe50495.png)


hope it helps!

erano067:
After some more investigation multer uses busboy version ^0.2.11 which gives 0.2.14.
[...]

PR #1092, opened 2022-05-17
#790 bump busboy up to solve encoding bug

erano067:
busboy in package json version(^0.2.11 = 0.2.14) does not handle multi fields in content-disposition.
so i bump up the version and fix all minor api breaking changes of busboy.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -30,7 +30,7 @@ function makeMiddleware (setup) {
     var busboy
 
     try {
-      busboy = new Busboy({ headers: req.headers, limits: limits, preservePath: preservePath })
+      busboy = Busboy({ headers: req.headers, limits: limits, preservePath: preservePath })
     } catch (err) {
       return next(err)
     }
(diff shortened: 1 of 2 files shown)


justinnitz:
A CVE is now associated with the dicer package. Upgrading to at least 1.0.0 of busboy removes dicer from multer's transitive dependency list. 

https://github.com/mscdex/busboy/issues/250 
https://security.snyk.io/vuln/SNYK-JS-DICER-2311764

mrded:
Two problems here:

1. The change is not formated: do `./node_modules/.bin/standard --fix`
2. The tests are failing: `npm run test` 
 - `Error Handling: should report errors from busboy parsing`
 -  `Unicode: 2) should handle unicode filenames`

mrded:
Two problems here:

1. The change is not formated: do `./node_modules/.bin/standard --fix`
2. The tests are failing: `npm run test` 
 - `Error Handling: should report errors from busboy parsing`
 -  `Unicode: 2) should handle unicode filenames`

progsmile:
> Two problems here
I've run format command on different node versions (8, 10, 12, 14, 16, 18) - it's ok.
Tests are failing on nodes: `v8.17.0`, `v9.11.2`, Instant error because of `catch` without error parameter.
Starting from node `v10.24.1` tests are ok from my local pc.

I see node.js 6 - is not fittable any more, because `standard` requires  node `>=8.10`. And travis.ci has many old versions.
What is requirement for node.js version?

progsmile:
> Two problems here:

What the version of node.js?

RopoMen:
@mrded unicode issue solution is in this PR https://github.com/expressjs/multer/pull/1102

LinusU:
This is done and discussed in #1097, currently it's published as `1.4.5-lts.1`

Issue #350, opened 2016-06-02
Error: Unexpected end of multipart data


soakit:
Error: Unexpected end of multipart data
    at C:\Users\admin\Desktop\wap\static\node_modules\dicer\lib\Dicer.js:62:28
    at nextTickCallbackWith0Args (node.js:420:9)
    at process._tickDomainCallback (node.js:390:13)

Server
> var upload = multer();
> router.post('/foo', upload.single('file'), function (req, res, next) {
>     console.log('foo');
> });

client:

> ClientRequest {                                                                                                                                                     
>   domain: null,                                                                                                                                                     
>   _events:                                                                                                                                                          
>    { response: { [Function: g] listener: [Function] },                                                                                                              
>      socket: { [Function: g] listener: [Function] } },                                                                                                              
>   _eventsCount: 2,                                                                                                                                                  
>   _maxListeners: undefined,                                                                                                                                         
>   output:                                                                                                                                                           
>    [ 'POST /foo HTTP/1.1\r\nContent-Type: multipart/form-data; boundary=-----np0.6452500836458057\r\nContent-Length: 286\r\nHost: localhost:8009\
> r\nConnection: close\r\n\r\n-------np0.6452500836458057\r\n',                                                                                                       
>      'Content-Disposition: form-data; name="to"\r\n',                                                                                                               
>      '\r\n',                                                                                                                                                        
>      '/data/static/admin/160601index/html/哈哈.html\r\n',                                                                                                       
>      '-------np0.6452500836458057\r\n',                                                                                                                             
>      'Content-Disposition: form-data; name="file"; filename="/html/哈哈.html"\r\n',                                                                                   
>      '\r\n',                                                                                                                                                        
>      <Buffer 61 61 61 0d 0a 62 62 62 0d 0a 61 61 61 0d 0a 61 62 63 0d 0a 61 73 73 61 0d 0a 73 73>,                                                                  
>      '\r\n',                                                                                                                                                        
>      '-------np0.6452500836458057--\r\n' ],                                                                                                                         
>   outputEncodings:                                                                                                                                                  
>    [ undefined,                                                                                                                                                     
>      undefined,                                                                                                                                                     
>      undefined,                                                                                                                                                     
>      undefined,                                                                                                                                                     
>      undefined,                                                                                                                                                     
>      undefined,                                                                                                                                                     
>      undefined,                                                                                                                                                     
>      undefined,                                                                                                                                                     
>      undefined,                                                                                                                                                     
[...]

PR #1094, opened 2022-05-20
fix: app crash

Sceri:
During stream termination, errors can occur which can crash the entire application, since all the listeners on the busboy are removed. We suppress such errors with this fix. 

Scenario: If Dicer didn't fully parse the content and the stream terminates (for example after failing to pass the fileFilter), Dicer sends an error event ([link](https://github.com/mscdex/dicer/blob/master/lib/Dicer.js#L68)). This event goes to Busboy ([link](https://github.com/mscdex/busboy/blob/v0.2.14/lib/types/multipart.js#L281)) and then crashes because all listeners have been deleted. 

This will potentially fix it: [1](https://github.com/expressjs/multer/issues/350) [2](https://github.com/nestjs/nest/issues/9489)

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -49,6 +49,7 @@ function makeMiddleware (setup) {
       req.unpipe(busboy)
       drainStream(req)
       busboy.removeAllListeners()
+      busboy.on('error', function (err) { })
 
       onFinished(req, function () { next(err) })
     }


LinusU:
Thanks for taking the time to work on this PR!

I recently released `1.4.4-lts.1` from the #1097 PR which removes Dicer altogether. Does that solve your problem?

ctcpip:
fixed in 2.0.0

PR #1096, opened 2022-05-20
bump busboy up to solve DoS vulnerability

mrded:
Hello,

Snyk is reporting a vulnerability in this repo, that is coming from the Dicer library:

```
Issues with no direct upgrade or patch:
  ✗ Denial of Service (DoS) [High Severity][https://snyk.io/vuln/SNYK-JS-DICER-2311764] in dicer@0.2.5
    introduced by multer@1.4.4 > busboy@0.2.14 > dicer@0.2.5
  No upgrade or patch available
```

This change removes dicer from multer's transitive dependency list.

PR #1056, opened 2021-12-04
Replace unsupported busboy version with fastify fork

kibertoad:

PR #1097, opened 2022-05-21
fix(cve): bump busboy to fix CVE-2022-24434

jlourenc:
This PR bumps `busboy` to at least `1.0.0` to remove `dicer` from the transitive dependencies as it contains a denial of service vulnerability: https://security.snyk.io/vuln/SNYK-JS-DICER-2311764.

The remaining of the PR is about adapting the code to the breaking changes introduced in `busboy` with the bump to `v1` and fixing the tests:
- reverted https://github.com/expressjs/multer/pull/205 to fix `should report errors from busboy parsing` test as the error was no longer forwarded. I have not really looked into the details but reverting this workaround for an issue related to Node v0 seems like a win anyway,
- updated `should handle unicode filenames` test to have a more pertinent example of a unicode filename using both `filename` and `filename*` `Content-Disposition` directives.

This PR supersedes: https://github.com/expressjs/multer/pull/1096, https://github.com/expressjs/multer/pull/1092 and https://github.com/expressjs/multer/pull/1056.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -1,17 +1,12 @@
 var is = require('type-is')
 var Busboy = require('busboy')
 var extend = require('xtend')
-var onFinished = require('on-finished')
 var appendField = require('append-field')
 
 var Counter = require('./counter')
 var MulterError = require('./multer-error')
 var FileAppender = require('./file-appender')
 var removeUploadedFiles = require('./remove-uploaded-files')
-
-function drainStream (stream) {
-  stream.on('readable', stream.read.bind(stream))
-}
 
 function makeMiddleware (setup) {
   return function multerMiddleware (req, res, next) {
@@ -30,7 +25,7 @@ function makeMiddleware (setup) {
     var busboy
 
     try {
-      busboy = new Busboy({ headers: req.headers, limits: limits, preservePath: preservePath })
+      busboy = Busboy({ headers: req.headers, limits: limits, preservePath: preservePath })
     } catch (err) {
       return next(err)
     }
@@ -45,12 +40,9 @@ function makeMiddleware (setup) {
     function done (err) {
       if (isDone) return
       isDone = true
-
       req.unpipe(busboy)
-      drainStream(req)
       busboy.removeAllListeners()
-
-      onFinished(req, function () { next(err) })
+      next(err)
     }
 
     function indicateDone () {
@@ -80,9 +72,9 @@ function makeMiddleware (setup) {
     }
 
     // handle text field data
-    busboy.on('field', function (fieldname, value, fieldnameTruncated, valueTruncated) {
+    busboy.on('field', function (fieldname, value, { nameTruncated, valueTruncated }) {
       if (fieldname == null) return abortWithCode('MISSING_FIELD_NAME')
-      if (fieldnameTruncated) return abortWithCode('LIMIT_FIELD_KEY')
+      if (nameTruncated) return abortWithCode('LIMIT_FIELD_KEY')
       if (valueTruncated) return abortWithCode('LIMIT_FIELD_VALUE', fieldname)
 
       // Work around bug in Busboy (https://github.com/mscdex/busboy/issues/6)
@@ -94,7 +86,7 @@ function makeMiddleware (setup) {
     })
 
     // handle files
-    busboy.on('file', function (fieldname, fileStream, filename, encoding, mimetype) {
+    busboy.on('file', function (fieldname, fileStream, { filename, encoding, mimeType }) {
       // don't attach to the files object, if there is no file
       if (!filename) return fileStream.resume()
 
@@ -107,7 +99,7 @@ function makeMiddleware (setup) {
         fieldname: fieldname,
         originalname: filename,
         encoding: encoding,
-        mimetype: mimetype
+        mimetype: mimeType
       }
 
       var placeholder = appender.insertPlaceholder(file)
@@ -169,7 +161,7 @@ function makeMiddleware (setup) {
     busboy.on('partsLimit', function () { abortWithCode('LIMIT_PART_COUNT') })
     busboy.on('filesLimit', function () { abortWithCode('LIMIT_FILE_COUNT') })
     busboy.on('fieldsLimit', function () { abortWithCode('LIMIT_FIELD_COUNT') })
-    busboy.on('finish', function () {
+    busboy.on('close', function () {
       readFinished = true
       indicateDone()
     })
--- a/package.json
+++ b/package.json
@@ -20,11 +20,10 @@
   ],
   "dependencies": {
     "append-field": "^1.0.0",
-    "busboy": "^0.2.11",
+    "busboy": "^1.0.0",
     "concat-stream": "^1.5.2",
     "mkdirp": "^0.5.4",
     "object-assign": "^4.1.1",
-    "on-finished": "^2.3.0",
     "type-is": "^1.6.4",
     "xtend": "^4.0.0"
   },
@@ -39,7 +38,7 @@
     "testdata-w3c-json-form": "^1.0.0"
   },
   "engines": {
-    "node": ">= 0.10.0"
+    "node": ">= 6.0.0"
   },
   "files": [
     "LICENSE",
--- a/test/_util.js
+++ b/test/_util.js
@@ -1,7 +1,6 @@
 var fs = require('fs')
 var path = require('path')
 var stream = require('stream')
-var onFinished = require('on-finished')
 
 exports.file = function file (name) {
   return fs.createReadStream(path.join(__dirname, 'files', name))
@@ -17,11 +16,6 @@ exports.submitForm = function submitForm (multer, form, cb) {
 
     var req = new stream.PassThrough()
 
-    req.complete = false
-    form.once('end', function () {
-      req.complete = true
-    })
-
     form.pipe(req)
     req.headers = {
       'content-type': 'multipart/form-data; boundary=' + form.getBoundary(),
@@ -29,7 +23,7 @@ exports.submitForm = function submitForm (multer, form, cb) {
     }
 
     multer(req, null, function (err) {
-      onFinished(req, function () { cb(err, req) })
+      cb(err, req)
     })
   })
 }
--- a/test/error-handling.js
+++ b/test/error-handling.js
@@ -244,7 +244,7 @@ describe('Error Handling', function () {
     req.end(body)
 
     upload(req, null, function (err) {
-      assert.strictEqual(err.message, 'Unexpected end of multipart data')
+      assert.strictEqual(err.message, 'Unexpected end of form')
       done()
     })
   })
--- a/test/express-integration.js
+++ b/test/express-integration.js
@@ -8,7 +8,6 @@ var util = require('./_util')
 var express = require('express')
 var FormData = require('form-data')
 var concat = require('concat-stream')
-var onFinished = require('on-finished')
 
 var port = 34279
 
@@ -27,7 +26,7 @@ describe('Express Integration', function () {
     req.on('response', function (res) {
       res.on('error', cb)
       res.pipe(concat({ encoding: 'buffer' }, function (body) {
-        onFinished(req, function () { cb(null, res, body) })
+        cb(null, res, body)
       }))
     })
   }
--- a/test/unicode.js
+++ b/test/unicode.js
@@ -2,12 +2,10 @@
 
 var assert = require('assert')
 
-var path = require('path')
-var util = require('./_util')
 var multer = require('../')
 var temp = require('fs-temp')
 var rimraf = require('rimraf')
-var FormData = require('form-data')
+var stream = require('stream')
 
 describe('Unicode', function () {
   var uploadDir, upload
@@ -34,21 +32,29 @@ describe('Unicode', function () {
   })
 
   it('should handle unicode filenames', function (done) {
-    var form = new FormData()
-    var parser = upload.single('small0')
-    var filename = '\ud83d\udca9.dat'
+    var req = new stream.PassThrough()
+    var boundary = 'AaB03x'
+    var body = [
+      '--' + boundary,
+      'Content-Disposition: form-data; name="small0"; filename="poo.dat"; filename*=utf-8\'\'%F0%9F%92%A9.dat',
+      'Content-Type: text/plain',
+      '',
+      'test with unicode filename',
+      '--' + boundary + '--'
+    ].join('\r\n')
 
-    form.append('small0', util.file('small0.dat'), { filename: filename })
+    req.headers = {
+      'content-type': 'multipart/form-data; boundary=' + boundary,
+      'content-length': body.length
+    }
 
-    util.submitForm(parser, form, function (err, req) {
+    req.end(body)
+
+    upload.single('small0')(req, null, function (err) {
       assert.ifError(err)
 
-      assert.strictEqual(path.basename(req.file.path), filename)
-      assert.strictEqual(req.file.originalname, filename)
-
+      assert.strictEqual(req.file.originalname, '\ud83d\udca9.dat')
       assert.strictEqual(req.file.fieldname, 'small0')
-      assert.strictEqual(req.file.size, 1778)
-      assert.strictEqual(util.fileSize(req.file.path), 1778)
 
       done()
     })


vaibhavkumar-sf:
This should be merged asap

LinusU:
As far as I'm aware, this would be a breaking change, right? I think we need to do this in 2.x instead?

jlourenc:
If the aim is to keep supporting Node v0 which is end-of-life since 2016, then yes it is a breaking change. However, such an aim is, in my opinion, counter-productive. Unless you're considering something else a breaking change?

Also, I've tried and run `npm run test` using `node v0.12.12` and the project is already not compatible with Node v0 I believe:
```
$ node -v
v0.12.12

$ npm run test

> multer@1.4.4 test /Users/jeremylourenco/Code/multer
> mocha --harmony

/Users/jeremylourenco/Code/multer/node_modules/busboy/lib/index.js:3
const { readdirSync } = require('fs');
      ^
SyntaxError: Unexpected token {
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/jeremylourenco/Code/multer/lib/make-middleware.js:2:14)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/jeremylourenco/Code/multer/index.js:1:84)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/jeremylourenco/Code/multer/test/disk-storage.js:9:14)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at /Users/jeremylourenco/Code/multer/node_modules/mocha/lib/mocha.js:231:27
    at Array.forEach (native)
    at Mocha.loadFiles (/Users/jeremylourenco/Code/multer/node_modules/mocha/lib/mocha.js:228:14)
    at Mocha.run (/Users/jeremylourenco/Code/multer/node_modules/mocha/lib/mocha.js:514:10)
    at Object.<anonymous> (/Users/jeremylourenco/Code/multer/node_modules/mocha/bin/_mocha:480:18)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
    at node.js:814:3
```

I'm not a Node expert but it seems "Object Destructuring" is a v6+ feature.

erano067:
Hi, why not update to busboy atleast version ^1.6. 0?

jlourenc:
> Hi, why not update to busboy atleast version ^1.6. 0?

The aim of this PR is to remove `dicer` from the dependencies, which is achieved by upgrading to at least `v1.0.0`. There is no reason to enforce a higher version as far as this PR is concerned. You may chose to enforce a higher version in your project pulling `multer`.

jlourenc:
I have now enforced the node engine to be `>= 6.0.0`.

vaibhavkumar-sf:
Hi, Thanks for all the updates, can you please bump the busboy version to the latest 1.6.0 otherwise we may need to upgrade it again sooner for other vulnerabilities in the near future?

harshagarwal00:
im stuck on this as well

ansonallard:
+1 on getting this merged in.

vaibhavkumar-sf:
Yes @Bozzzzzo , thanks, I didn't notice that it is ^ instead of ~ 
@LinusU  please merge it

achilehero:
+1 on merging and releasing

kutluturk:
+1 on merging. This is a serious security problem

dizhenlyu:
+1 on the merge and release ASAP. Thank you!

zacharytyhacz:
```
 _______________________ 
< let's get this merged >
 ----------------------- 
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||


```

eran10:
+1,can this please be merge

senpai-notices:
lgtm. i approve this mr. do let any of us know if there are any blockers.

zSakuraEvilz:
+1, can this be merged and release v1.4.5 please ?

MohammedHMCTS (approved):

progsmile:
Let's release bro :)

NormandoHall:
c'mon!

gabrielrra (approved):

ministere-famille-qc:
> This PR bumps `busboy` to at least `1.0.0` to remove `dicer` from the transitive dependencies as it contains a denial of service vulnerability: https://security.snyk.io/vuln/SNYK-JS-DICER-2311764.
> 
> The remaining of the PR is about adapting the code to the breaking changes introduced in `busboy` with the bump to `v1` and fixing the tests:
> 
> * reverted [lib: work around node.js bug #205](https://github.com/expressjs/multer/pull/205) to fix `should report errors from busboy parsing` test as the error was no longer forwarded. I have not really looked into the details but reverting this workaround for an issue related to Node v0 seems like a win anyway,
> * updated `should handle unicode filenames` test to have a more pertinent example of a unicode filename using both `filename` and `filename*` `Content-Disposition` directives.
> 
> This PR supersedes: #1096, #1092 and #1056.

My NestJs app doesn't compile, so i found this temporary fix: 

As of npm cli [v8.3.0](https://github.com/npm/cli/releases/tag/v8.3.0) (2021-12-09) this can be solved using the [overrides field](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) of package.json

**For NestJs : ** 
`overrides": {
    "@nestjs/platform-express": {
      "multer": {
        "busboy": "1.0.0"
`
**For Express :** 
`overrides": {
      "multer": {
        "busboy": "1.0.0"
`

vaibhavkumar-sf:
override is a temporary solution to use until the package owner does not upgrade the package, also override is okay only if you are sure that the required version will not break the package @Dany-C

LinusU:
Hey everyone!

I understand that this is frustrating for everyone involved. Unfortunately, I also see it as a big problem to release this as 1.x since it would break compatibility with older Node.js versions. Older versions of Node.js is unfortunately still used in many production environment, and it's Express policy to not break Node.js compatibility in non-major changes.

It also doesn't make sense to release this as `2.x` since we already have several release candidates of Multer 2.x already, and they include an all new API.

Btw. I have released version `2.0.0-rc.4` which uses `@fastify/busboy` instead, and thus shouldn't be vulnerable.

I would love to get some input on how to proceed, some alternatives that I see:

1. Publish this branch as e.g. `multer-classic` on Npm
2. Publish this branch as something like `1.4.5-node.12`, which I think won't get picked up as automatic update in semver?
3. Release this as `1.4.5` as a breaking change
4. Fork dicer (& busboy) and just fix the CVE, but keep Node.js compatibility
5. Inline the code for busboy and then restore old Node.js compatibility to it

@dougwilson do you have any input or guidance?

cnorthwood:
An alternate to consider- release as version 3.x and then the current 2.x RCs when ready for release are released as version 4? This would be semver compatible but feels a bit nasty to then essentially "skip" a version but that is an aesthetic consideration imo so the need for pragmatism might overrule it.

jlourenc:
Hey @LinusU,

> Unfortunately, I also see it as a big problem to release this as 1.x since it would break compatibility with older Node.js versions.

I've tested again compatibility with node locally and yes, you're right this is a breaking change.

My 2 cents would be to:
- use option 2 if that does not get picked up automatically
- mark all other `v1` versions as deprecated to encourage people moving off of them, to the `v1.4.5-node.12` in the short term but to the `v2` in the long term
- stop `v1` development and only accept security patches while `v2` gets released

zSakuraEvilz:
So is there any way to use v1 without security? Because the number of sources is quite large, if you upgrade to v2, you have to modify the current code, which is very difficult for big projects.

rudxde:
Hi,
i think if you would publish the fix first under the `1.4.5-node.12` tag, most user would be able to fix this issue in their application, since upgrading to any `v2-rc` is not possible for every project and will leave many applications longer vulnerable. Afterwards it's still possible to work out, how to continue on this issue.

vaibhavkumar-sf:
voting for 1.4.5-node.12

mathertel (approved):

georgerdp-cic:
> MUST rate security issues higher than compatibility. fixing is forward only. @jlourenc did a good job. install using
> 
> ```
> npm i "jlourenc/multer.git#fix/CVE-2022-24434"
> ```
> 
> and test before complaining. I had no problem for my applications.

Nice work.  I've uploaded successfully multiple files.
When I attempt to use the malicious code now I get this error.

``` 
Error message: Malformed part header
```

```
Code to test:

 fetch('<- upload endpoint ->', {
      method: 'POST',
      headers: {
        ['content-type']: 'multipart/form-data; boundary=----WebKitFormBoundaryoo6vortfDzBsDiro',
        ['content-length']: '145',
        connection: 'keep-alive',
      },
      body: '------WebKitFormBoundaryoo6vortfDzBsDiro\r\n Content-Disposition: form-data; name="bildbeschreibung"\r\n\r\n\r\n------WebKitFormBoundaryoo6vortfDzBsDiro--'
    });

inPhoenix:
つ ◕＿◕ ༽つ つ ◕＿◕ ༽つつ ◕＿◕ ༽つ +1

rudxde (approved):

benasher44:
Option 4 would both avoid breaking changes here and provide an easy upgrade path for others that can't get off of dicer yet

LinusU:
I have now published version `1.4.4-lts.1` to npm, it's also under the `latest` tag which means running `npm install multer` in a fresh project should pick that version.

Also, I have deprecated all other 1.x versions with the following message:

> Multer 1.x is affected by CVE-2022-24434. This is fixed in v1.4.4-lts.1 which drops support for versions of Node.js before 6. Please upgrade to at least Node.js 6 and version 1.4.4-lts.1 of Multer. If you need support for older versions of Node.js, we are open to accepting patches that would fix the CVE on the main 1.x release line, whilst maintaining compatibility with Node.js 0.10.

harshagarwal00:
@LinusU : i dont see this ?
maybe i missed something..
https://www.npmjs.com/package/multer
https://github.com/expressjs/multer/tags

LinusU:
@harshagarwal00 it shows for me on npm, you might be getting a cached version, will probably be updated soon. Have you tried running `npm install multer`?

Screenshot 2022-05-29 at 09 40 22

It did forget to push with `--follow-tags` though, will fix that now 😅

harshagarwal00:
@LinusU : it works brilliant :D 
thank you !

mpsijm:
Thanks for the fix! However, it appears that the tag `1.4.4-lts.1` is not "newer" than `1.4.4` according to semver: https://semvercompare.azurewebsites.net/?version=1.4.4-lts.1&version=1.4.4&version=1.4.3

I had `"multer": "^1.4.4-lts.1"` in my `package.json`, and it installed the old version `1.4.4` instead, due to this semver ordering.

Removing the `^` from the package version or manually running `npm i multer@1.4.4-lts.1` seemed to fix this, but that is undesirable.

ghost91-:
What @mpsijm said is correct, and in particular, this is problematic for libraries that have been using a version range, using `^` (i.e. something like `"multer": "^1.4.4")`, because it means you can't easily use the correct version in packages that use libraries using multer. _Al_l those libraries would need to be updated to to pin the version of multer to `1.4.4-lts.1`.

Is there a chance of releasing this as 1.4.5 after all? Otherwise, this will take quite a bit of time until it actually trickles down into all apps. And it may not even be possible to resolve in situations where libraries aren't currently maintained, so nobody can pin the version...

LinusU:
I've published version `1.4.5-lts.1` without any changes, in order to fix the problem that `1.4.4` was considered higher than `1.4.4-lts.1`.

> Is there a chance of releasing this as 1.4.5 after all? Otherwise, this will take quite a bit of time until it actually trickles down into all apps. And it may not even be possible to resolve in situations where libraries aren't currently maintained, so nobody can pin the version...

I don't see that we can release a `1.4.5` without the `-lts.*` postfix unless someone contributes a fix that works without breaking older versions of Node.js...

Bozzzzzo:
> I've published version `1.4.5-lts.1` without any changes, in order to fix the problem that `1.4.4` was considered higher than `1.4.4-lts.1`.

Thanks ... but It does not help as "1.4.5-lts.1" does not match semver like "^1.0.0" either.

> I don't see that we can release a `1.4.5` without the `-lts.*` postfix unless someone contributes a fix that works without breaking older versions of Node.js...

Are you sure you really want to support such old versions of NodeJS ? Version before v6 were published before 2016 !
In fact even v6 is obsolete and its support ended for quite a long time!

Please have a look at Node.JS long-term-support schedule at : https://nodejs.org/en/about/releases/

Personally I don't think you would do anyone a favor by trying to support such old versions of Node.JS run time(s)
as such old runtime(s) did not get even security fixes for quite a long time and will never;
 they are full of dreadful vulnerabilities!
 it does not make sense to incite people to keep developing on such obsolete version.

Besides as ghost91 ; I am also in a situation where I cannot apply 1.4.5-lts.1
(I can't "override" either as I am stuck on NodeJS v14 because of some other dependencies)

mpsijm:
> Are you sure you really want to support such old versions of NodeJS ?

I do agree with you that old versions of NodeJS should no longer be supported. However, an important point of SemVer is that minor releases must not introduce backwards incompatible changes. Even through those old versions of NodeJS are long not supported anymore, version 1.x will have to support them forever, according to SemVer.

One possible solution would be to release this backwards-incompatible fix as version 2.0.0. At least, that's what I see when other packages drop certain Node versions: they bump the major version number, even though the rest of the changes are not super significant. A problem that I see with this however, is that `multer` has been publishing pre-releases for `2.0.0` for some years[1], probably _actually_ containing new features, and simply publishing this seemingly minor change as 2.0.0 may lead to confusion as well.

[1] Source: "Version" tab at https://www.npmjs.com/package/multer

EstartuPrime:
@LinusU You're right, we updated to 1.4.5-lts.1 manually as we noticed the deprecation information and read in youre deprication wraning, that this 1.4.5-lts.1 still supports Node 6. But during startup of the application it crashes when busboy is loaded by the reason i linked to busboy.

We switched back to 1.4.4 and the application is back to normal. 

All this took me some time to find out where exactly the problem is and if we can solve it anyway and can keep the security fix. 
And as i understand in the linked Question, this "issue" will not be fixed in Busboy.
That's why i thought i better informed you about this incompatibility with Node.js 6 and busboy 1.6.0 , so that you may update your'e deprecation warning.

RopoMen:
Hi,
I think that better fix to unicode issue can be achieved through this PR https://github.com/expressjs/multer/pull/1102

zoeesilcock:
Another package that is affected by this is `multer-gridfs-storage`, see this [issue](https://github.com/devconcept/multer-gridfs-storage/issues/490). That package already requires Node 12 so it shouldn't require a major step for them. I would like to provide a PR for that package, am I correct in thinking that the solution for packages in this situation is to change their peer dependecy so it is pinned to `1.4.5-lts.1`? Or would `^1.4.5-lts.1` work to allow future versions?

ghost91-:
The latter should work, too.

alasdairhurst:
From what i can see, busboy 1.0.0+ has documented a requirement of node >=10.16.0. Is this an oversight or is there a reason this is being ignored and the engines in multer being set to a lower version (6)?

hongyanh:
`npm uninstall --save multer` and `npm install --save multer` fixed my problem.

UlisesGascon:
This patch was included in multer@1.4.4 and the engines discussion is continuing in https://github.com/expressjs/multer/pull/1306

Issue #846, opened 2020-02-25
'preservePath:true' is very unclear


hanani8:
I wanted to fetch the full path of a file that's being uploaded by the user. I have found about the preservePath option, I have enabled it. But the documentation is so unclear, it's very rudimentary to figure out, where to go from here

jonchurch:
You're right the docs are very unclear about that particular option.

Would you consider creating a PR to improve them?

Issue #1082, opened 2022-03-30
originalfilename error code when original contain chinese


realmxu:
w�1990.jpeg

realmxu:
{
  fieldname: 'file',
  originalname: 'w�1990.jpeg',
  encoding: '7bit',
  mimetype: 'image/jpeg',
  destination: './upload',
  filename: 'd93f5b48-e576-4f76-97e3-bd41bc3f7ef0.jpeg',
  path: 'upload\\d93f5b48-e576-4f76-97e3-bd41bc3f7ef0.jpeg',
  size: 189400
}

kimkanghyune:
{
    fieldname: 'media',
    originalname: 'Tt �� 2022-05-22 130102.jpg',
    encoding: '7bit',
    mimetype: 'image/jpeg',
    buffer: <Buffer ff d8 ff e0 00 10 4a 46 49 46 00 01 01 01 00 60 00 60 00 00 ff db 00 43 00 03 02 02 03 02 02 03 03 03 03 04 03 03 04 05 08 05 05 04 04 05 0a 07 07 06 ... 810242 more bytes>,
    size: 810292
  },

I also have the same problem

PR #1102, opened 2022-06-02
Added support for  busboy configuration parameters; 'defCharset', 'defParamCharset', 'highWaterMark' and 'fileHwm'

RopoMen:
Hi,

I added support for these busboy configuration paramters. Reason for this is that I need ' defParamCharset' override from `latin1` to `utf-8`. We have busboy 1.6.0 in our project and busboy has changed filename parsing from utf-8 to latin1 and it broke our application. Our Multer version is `1.4.5-lts.1` so these changes should be release on the lts side as well.

I would not like to add `iconv-lite` package to our application, I would prefer Multer to allow to set these properties.

I tried to upgrade busboy to 1.6.0 in Multer, but it broke too many unit tests (30+). I changed `test/unicode.js` to use filename `åäöèøßð.dat` and with busboy 1.6.0 result will be `aÌaÌoÌeÌÃ¸ÃÃ°.dat` and it can be fixed by adding busboy configuration `defParamCharset = 'utf-8'`

Fixes:
#320 
#846

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,10 +135,14 @@ The following are the options that can be passed to Multer.
 
 Key | Description
 --- | ---
-`dest` or `storage` | Where to store the files
-`fileFilter` | Function to control which files are accepted
-`limits` | Limits of the uploaded data
-`preservePath` | Keep the full path of files instead of just the base name
+`dest` or `storage` | Where to store the files.
+`fileFilter` | Function to control which files are accepted.
+`limits` | Limits of the uploaded data. **Defaults:** check [busboy's page](https://github.com/mscdex/busboy#exports)
+`preservePath` | Keep the full path of files instead of just the base name. **Default:** check [busboy's page](https://github.com/mscdex/busboy#exports)
+`charset` | Character set to use when one isn't defined. Sets busboy's option `defCharset`. **Default:** `utf-8`. List of [available charsets](https://github.com/mscdex/busboy/blob/master/lib/utils.js#L384)
+`paramCharset` | For multipart forms, the default character set to use for values of part header parameters (e.g. filename) that are not extended parameters (that contain an explicit charset). Sets busboy's option `defParamCharset`. **Default:** `utf-8`. List of [available charsets](https://github.com/mscdex/busboy/blob/master/lib/utils.js#L384).
+`writableHVM` | Node Stream highWaterMark to use for the parser stream. Sets busboy's option `highWaterMark`. **Default:** check [busboy's page](https://github.com/mscdex/busboy#exports)
+`readableHVM` | Node Stream highWaterMark to use for individual file streams. Sets busboy's option `fileHwm`. **Default:** check [busboy's page](https://github.com/mscdex/busboy#exports)
 
 In an average web app, only `dest` might be required, and configured as shown in
 the following example.
@@ -260,7 +264,7 @@ memory storage is used.
 
 ### `limits`
 
-An object specifying the size limits of the following optional properties. Multer passes this object into busboy directly, and the details of the properties can be found on [busboy's page](https://github.com/mscdex/busboy#busboy-methods).
+An object specifying the size limits of the following optional properties. Multer passes this object into busboy directly, and the details of the properties can be found on [busboy's page](https://github.com/mscdex/busboy#exports).
 
 The following integer values are available:
 
--- a/index.js
+++ b/index.js
@@ -18,6 +18,10 @@ function Multer (options) {
   }
 
   this.limits = options.limits
+  this.charset = typeof options.charset === 'string' ? options.charset : 'utf-8'
+  this.paramCharset = typeof options.paramCharset === 'string' ? options.paramCharset : 'utf-8'
+  this.writableHVM = options.writableHVM
+  this.readableHVM = options.readableHVM
   this.preservePath = options.preservePath
   this.fileFilter = options.fileFilter || allowAll
 }
@@ -48,6 +52,10 @@ Multer.prototype._makeMiddleware = function (fields, fileStrategy) {
       limits: this.limits,
       preservePath: this.preservePath,
       storage: this.storage,
+      charset: this.charset,
+      paramCharset: this.paramCharset,
+      writableHVM: this.writableHVM,
+      readableHVM: this.readableHVM,
       fileFilter: wrappedFileFilter,
       fileStrategy: fileStrategy
     }
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -24,13 +24,24 @@ function makeMiddleware (setup) {
     var fileFilter = options.fileFilter
     var fileStrategy = options.fileStrategy
     var preservePath = options.preservePath
-
+    var charset = options.charset
+    var paramCharset = options.paramCharset
+    var writableHVM = options.writableHVM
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -135,10 +135,14 @@ The following are the options that can be passed to Multer.
...
-`dest` or `storage` | Where to store the files
-`fileFilter` | Function to control which files are accepted
-`limits` | Limits of the uploaded data
-`preservePath` | Keep the full path of files instead of just the base name
+`dest` or `storage` | Where to store the files.
+`fileFilter` | Function to control which files are accepted.
+`limits` | Limits of the uploaded data. **Defaults:** check [busboy's page](https://github.com/mscdex/busboy#exports)
+`preservePath` | Keep the full path of files instead of just the base name. **Default:** check [busboy's page](https://github.com/mscdex/busboy#exports)
+`charset` | Character set to use when one isn't defined. Sets busboy's option `defCharset`. **Default:** `utf-8`. List of [available charsets](https://github.com/mscdex/busboy/blob/master/lib/utils.js#L384)
+`paramCharset` | For multipart forms, the default character set to use for values of part header parameters (e.g. filename) that are not extended parameters (that contain an explicit charset). Sets busboy's option `defParamCharset`. **Default:** `utf-8`. List of [available charsets](https://github.com/mscdex/busboy/blob/master/lib/utils.js#L384).
```
zacharytyhacz, line 143: Shouldn't the default charset be `latin1` if the existing default is `latin1` ? 

You are trying to override `latin1` to be `utf-8` correct?

RopoMen: > You are trying to override latin1 to be utf-8 correct ( in your case ) ?

Yes, and the reason for that is that Busboy has parsed filenames as utf8 before this commit 3 months ago https://github.com/mscdex/busboy/commit/d7e4e2dfc4c98c72ec9d495b38ba38c2219bfe87

My opinion is that Multer should set its own defaults to ensure consistent behaviour and not rely on busboys defaults, because those may change without notice. This could be dealt with fixed version pinnings in package.json, but that may not be the best option.

Also `preservePath` should be set to `false` by Multer to ensure consistent behaviour. Although with that parameter I don't think that Busboy would change it to be `true`. Reasoning for this is that it depend on the client whether or not it sends the path of the file or not. And most of the clients don't send it.

And for the parameter handling; I didn't what to use same parameter names as Busboy because I see that Multer could someday relay on some other solution than Busboy and for example `highwatermark` params were named inconsistently in Busboy.

RopoMen: Busboy currently is not interested if `filename` field is said to be utf-8, it still uses `latin1` decoder at least in Busboy 1.6.0.

Here Busboy [takes filename](https://github.com/mscdex/busboy/blob/master/lib/types/multipart.js#L316-L323) from multipart form. It does not matter whether or not `filename` has encoding set like this `filename*=utf-8` it [still uses same decoder](https://github.com/mscdex/busboy/blob/master/lib/types/multipart.js#L305-L306), that is [defined here](https://github.com/mscdex/busboy/blob/master/lib/types/multipart.js#L236-L239) and if `defParamCharset` is not given, decoder will be `latin1`

And because of this knowledge, I think that this unit test in `lts` branch is broken https://github.com/expressjs/multer/blob/lts/test/unicode.js#L37-L44 and proper fix would be configuring `defParamCharset: 'utf8'` into Busboy, either Multer provides configuration option for that charset or hardcodes one into [make-middleware.js](https://github.com/expressjs/multer/blob/lts/lib/make-middleware.js#L28)

We also tested yesterday that Chrome is not adding encoding like this `filename*=utf-8` into form. (we are using `FormData`)

I just checked and Busboy 1.4.0 has worked ok with utf-8 filenames, but after that the breaking commit is added
Screenshot 2022-06-03 at 7 29 31

RopoMen: Perhaps I should comment this Unicode unit test again https://github.com/expressjs/multer/blob/lts/test/unicode.js#L37-L44

Because Busboy is handling `filename*=utf-8` in a same way as `filename=` AND because Busboy >=1.4.0 parsed filenames always as utf-8; that test passed always. With Busboy 1.5.0+ that test should fail.

RopoMen: Tested, it didn't fail. So I'm missing now something here. 🤔 
  
We had Multer 1.4.2 that installed Busboy 0.2.14. Filename `test-åäöèøßð.pdf` is uploaded from Chrome 100
Screenshot 2022-06-03 at 9 45 23

Filename looks like that and Multer returns `req.file.originalname` as `test-åäöèøßð.pdf`

Because of this npm audit issue https://github.com/advisories/GHSA-wm7h-9275-46v2 we updated Multer to 1.4.5-lts.1 that installs Busboy 1.6.0

Now, browser sends file as it did previosly `test-åäöèøßð.pdf`, but Multer `req.file.originalname` is `test-aÌaÌoÌeÌÃ¸ÃÃ°.pdf`. This can be fixed by setting `defParamCharset = utf-8` into Busboy.

RopoMen:
@zacharytyhacz can you give any estimate how quickly this could be merged, cherry-picked into `lts` and making new LTS release? If ETA is 1-3 days I can wait, otherwise I need to add `iconv-lite` and fix this temporarily.

RopoMen:
@LinusU ping.

RopoMen:
Also issue #1082 is related to this.

RSS1102:
I also encountered the problem of "Chinese garbled code" today. Thank you. I got the solution - v1.4.5-lts 1.thanks.

RopoMen:
I now use this temporary fix  ( https://github.com/expressjs/multer#multeropts )

```
fileFilter: (req, file, cb) => {
  file.originalname = Buffer.from(file.originalname, 'latin1').toString(
    'utf8'
  )
  cb(null, true)
},
```

marckornberger:
@RopoMen After an entire day of eternal pain, you are truly a hero! Thanks so much for posting your solution! ❤️

adrianmxb:
It is a good idea to make this configurable.
Maybe someone can also add a notice to the recent release or something as the update from 1.44 to 1.45 is breaking stuff.

luoxzhg:
https://github.com/mscdex/busboy/issues/247

> I now use this temporary fix ( https://github.com/expressjs/multer#multeropts )
> 
> ```
> fileFilter: (req, file, cb) => {
>   file.originalname = Buffer.from(file.originalname, 'latin1').toString(
>     'utf8'
>   )
>   cb(null, true)
> },
> ```

luojiong:
This problem still exists until October 22  @RopoMen  but thanks    I solved this problem with the method you gave me。my version is 1.4.7

RopoMen:
@Flo0806 check this npm package https://www.npmjs.com/package/content-disposition

jhpung:
I published a `multer-utf8` on npm that read files as `utf8` charset by default.

https://www.npmjs.com/package/multer-utf8

mkesper:
How someone can think that latin1 is still preferrable to utf-8 is beyond me. Please merge a fix for this so we can use non-ASCII chars in filenames without [workarounds](https://github.com/expressjs/multer/pull/1102#issuecomment-1214736616).

Doc999tor:
This PR https://github.com/expressjs/multer/pull/1210 is backward compatible to the current `multer` behavior

UlisesGascon:
Superseded by #1465

Issue #814, opened 2019-12-12
CLS Context is lost after using multer middleware


josencv:
Hello! I'm using the [cls-hooked](https://www.npmjs.com/package/cls-hooked) library, which let's me use continuous local storage in my express app. After I use the `multer.file` middleware, my context is lost.

I patched the middleware using the solution proposed [here](https://github.com/othiym23/node-continuation-local-storage/issues/29#issuecomment-226965698) and now it works fine. I'm letting you know just in case you want to fix it.

Cheers,

**Additional info**
Node version: 10.15.3
Express version: 4.16.3
Multer version: 1.4.2
CLS Hooked version: 4.2.2
OS: Ubuntu 18.04.3 LTS

Tchoupinax:
Hello,

I have the same issue with the object `AsyncLocalStorage` (from `'async_hooks'`), the context is lost on the route which has the multer middleware.

grzegorzjudas:
I struggled for a long time with the same issue, but I was able to go around it by passing all necessary middlewares with the following notation:

```typescript
async function uploadMiddleware (req: Request, res: Response, next: NextFunction) {
    try {
        await new Promise<void>((resolve, reject) => {
            multer({
                // ...
            }).array('files')(req, res, (error: any) => {
                if (error) return reject(error);

                return resolve();
            });
        });

        next();
    } catch (error) {
        next(error);
    }
}

server.get('/endpoint', uploadMiddleware, controller);
```

So basically what I'm doing is, I'm running the multer middleware in-directly by wrapping it in another middleware that handles context correctly and then, at correct time, continuing to next middlewares/controller by calling `next()`.

ajhyndman:
I just thought I'd share a cheap user-space workaround that my team came up with, in case it's helpful for anyone else.

```js
const { AsyncResource } = require('node:async_hooks')

/**
 * Ensure that an express middleware respects async local storage context.
 *
 * Accepts an express middleware and returns a copy of that middleware with the
 * next() call explicitly bound to the current AsyncResource context.
 *
 * When working with native promises or async/await syntax, Node can
 * automatically forward any existing async resource to nested async or promise
 * invocations. However, when working with nested callback-passing patterns
 * (such as occurs in an express middleware stack) Node cannot know whether the
 * function being passed around should be bound to the current resource.
 *
 * Express may update their plugin architecture to resolve this in a future
 * version. Middleware developers can also update their implementations to
 * ensure backward-compatible support for AsyncLocalStorage. This function
[...]

PR #1119, opened 2022-07-07
fix: wrap middleware in promise

InsOpDe:
this is necessary, so that async hooks work

#814
#1046

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -14,7 +14,7 @@ function drainStream (stream) {
 }
 
 function makeMiddleware (setup) {
-  return function multerMiddleware (req, res, next) {
+  function multerMiddleware (req, res, next) {
     if (!is(req, ['multipart'])) return next()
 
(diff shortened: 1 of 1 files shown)


InsOpDe:
tests all pass.

no new tests have been added

LinusU:
Thank you for the PR! Currently there are unfortunately two problems which means we cannot merge it yet:

- This is a breaking change, since Multer 1.x supports Node.js 0.10 which doesn't have `Promise`
- If `next` throws an error that will be caught by us, and then we will call `next` again with that error. This is a very common misstake to make when trying to "callbackify" a promise. You can look here for some info on doing it the "correct" way: https://github.com/LinusU/util-callbackify/blob/master/index.js

UlisesGascon:
Closing in favour of #1124, which fixes the same async-context loss (#814, #1046) with an AsyncResource and without the double-next problem Linus pointed out. Thanks for raising it back then.

Issue #256, opened 2015-10-25
How can cancel upload during file is uploading ?


juniorkid:
I want to cancel upload during file is uploading .
How can i do it ?

LinusU:
I'm guessing that you want to cancel the upload from the client-side. It would depend on which library you are using.

`XMLHttpRequest` has an [`abort` function](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#abort()).

Feel free to reopen if you need more help.

PR #1120, opened 2022-07-10
fix: MEMORY LEAK

UnlimitedBytes:
One important caveat was ignored by piping request stream to busboy stream. This commit aknowledges that caveat by closing busboy stream when request stream encounters an error. More information about this caveat can be found on https://nodejs.org/api/stream.html#readablepipedestination-options.
> One important caveat is that if the Readable stream emits an error during processing, the Writable destination is not closed automatically. If an error occurs, it will be necessary to manually close each stream in order to prevent memory leaks.

This PR also fixes multiple issues that are caused by this caveat some are listed blow.
- #259
- #256
- #38
- (maybe) #53
- (maybe) #1058

Some PR's (that may introduce breaking changes) are closed by this PR also (without breaking changes) and some are listed blow.
- #1117
- #971
- #438
- #429

**PLEASE MERGE THIS BECAUSE IT'S INSANE ANNOYING HAVING TO RELAUNCH THE SERVER EVERY DAY BECAUSE THOUSANDS OF OPEN STREAMS ARE NOT CLOSED (MEMORY LEAKS)**

@LinusU here is a not breaking change that will fix all the above.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -42,10 +42,16 @@ function makeMiddleware (setup) {
     var pendingWrites = new Counter()
     var uploadedFiles = []
 
+    function requestError(error) {
+      if(!busboy) return
+      busboy.destroy(error)
+    }
+    
     function done (err) {
       if (isDone) return
       isDone = true
 
+      req.removeEventListener('error', requestError)
       req.unpipe(busboy)
       drainStream(req)
       busboy.removeAllListeners()
@@ -174,6 +180,7 @@ function makeMiddleware (setup) {
       indicateDone()
     })
 
+    req.on('error', requestError)
     req.pipe(busboy)
   }
 }
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -51,7 +51,7 @@ function makeMiddleware (setup) {
       if (isDone) return
       isDone = true
 
-      req.removeEventListener('error', requestError)
+      req.removeListener('error', requestError)
       req.unpipe(busboy)
       drainStream(req)
       busboy.removeAllListeners()


lib/make-middleware.js
```diff
@@ -42,10 +42,16 @@ function makeMiddleware (setup) {
...
+    function requestError(error) {
+      if(!busboy) return
+      busboy.destroy(error)
+    }
+    
     function done (err) {
       if (isDone) return
       isDone = true
 
+      req.removeEventListener('error', requestError)
```
Gerard-Szulc, line 54: removeEventListener gives me error, try to use req.off('error', requestError)

lib/make-middleware.js
```diff
@@ -42,10 +42,16 @@ function makeMiddleware (setup) {
     var pendingWrites = new Counter()
     var uploadedFiles = []
 
+    function requestError(error) {
+      if(!busboy) return
```
LinusU, line 46: `busboy` should always be defined here?

```suggestion
```

lib/make-middleware.js
```diff
@@ -42,10 +42,16 @@ function makeMiddleware (setup) {
     var pendingWrites = new Counter()
     var uploadedFiles = []
 
+    function requestError(error) {
```
LinusU, line 45:
```suggestion
    function requestError (error) {
```

LinusU:
Sorry for the late reply, this looks good 👍 

Only have one question, will the `next` be called with an error when this happens? Could you add a test case covering this?

r4881t:
@LinusU Merger?

iMrDJAi:
@LinusU Can you please merge this? Servers that use `multer` are exposed to DOS attacks!

ctcpip:
resolved via https://github.com/expressjs/multer/commit/2c8505f207d923dd8de13a9f93a4563e59933665

thanks everyone!

PR #1124, opened 2022-07-24
fix: preserve async hooks context in supported environments

### LINKED ISSUE #814: CLS Context is lost after using multer middleware

### LINKED ISSUE #1046: next() function not called in middleware, which breaks express-http-context (cls-hooked)

ajhyndman:
Hey, I took another attempt at fixing multer's support for `AsyncLocalStorage`.

I believe this addresses the feedback in #1119.

Fixes #814
Fixes #1046

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -15,6 +15,13 @@ function drainStream (stream) {
 
 function makeMiddleware (setup) {
   return function multerMiddleware (req, res, next) {
+    // If we are in a node environment that supports async_hooks, preserve
+    // the current async context.
+    try {
+      var asyncHooks = require('node:async_hooks')
+      next = asyncHooks.AsyncResource.bind(next)
+    } catch (_) {}
+
     if (!is(req, ['multipart'])) return next()
 
     var options = setup()
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -1,4 +1,5 @@
 var is = require('type-is')
+var AsyncResource = require('async_hooks').AsyncResource
 var Busboy = require('busboy')
 var appendField = require('append-field')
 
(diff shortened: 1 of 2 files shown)


saifabusaleh:
Hi, Any ETA when this will be merged?

Tyler8oliver:
Any ETA on release of this?

rsauget:
Hi @LinusU 👋 
We just experienced a production issue that would have been avoided with this PR, do you think we could merge it?

xseman:
Hi @LinusU @ajhyndman is there anything I can assist with to help move it forward?

MagicIndustries:
We're experiencing the same issue, any updates on getting it merged?

prosenkiewicz-tsh:
FYI: if anyone struggles with the issue, here is great solution for that https://github.com/expressjs/multer/issues/814#issuecomment-1218998366

seripjc:
@UlisesGascon this issue is still reproducible on the latest Multer version. What’s needed to move this forward and get a release out? Please let us know if there’s anything we can do to help. Otherwise, it would be great to get this merged, as it's been a known issue for over three years now.

UlisesGascon:
Sorry for the big delay, this goes into the next release!

Pushed a commit on top (212b3fd): switched to `AsyncResource#runInAsyncScope` (the static bind drops the argument on Node 14 and node: requires don't exist before 14.18, which made the test hang on the older legs) and made the test skip instead of try/catch when AsyncLocalStorage isn't available.

UlisesGascon (approved):

PR #1125, opened 2022-07-29
build: :arrow_up: use newer version of busboy to fix #1122

### LINKED ISSUE #1122: Security vulnerability due to dicer 0.2.5
Dear Team

In our product the **high security vulnerability** has been reported due to the nested sub-package dicer 0.2.5 even in the latest version (1.4.4) of multer. Would you please help to check and share your mitigation plan if it's planned.

image

Best Regards,
Hank.

pshaddel:
dicer has a security issue and busboy is not using it in the newer version so we need to update this package

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -30,7 +30,7 @@ function makeMiddleware (setup) {
     var busboy
 
     try {
-      busboy = new Busboy({ headers: req.headers, limits: limits, preservePath: preservePath })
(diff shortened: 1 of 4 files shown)


LinusU:
This is a duplicate of #1097.

The change have already been published as `1.4.5-lts.1` which drops support for older versions of Node.js.

We can only put this in a normal `1.x` release if someone can upgrade whilst keeping compatibility with Node.js 0.10.

PR #426, opened 2016-12-08
feat. improve wording for LIMIT_UNEXPECTED_FILE error code

flashbag:

PR #1128, opened 2022-08-09
fix: Edit LIMIT_UNEXPECTED_FILE error message to be Unexpected File.

bigboypantson:
Fixed error message for LIMIT_UNEXPECTED_FILE.


avinashkamat48 (commented):
`LIMIT_UNEXPECTED_FILE`'s message is part of the observable error object, so this may break users or tests that assert the previous exact text (`Unexpected field`). If the change is intentional, it would be good to add/update a test that locks in the new message and mention the compatibility impact in the changelog/release notes. Otherwise this might be safer as documentation only, because the code has historically used "field" for the form field name that triggered the error.

UlisesGascon:
Duplicate. see https://github.com/expressjs/multer/pull/426

PR #1133, opened 2022-08-28
feat: accept a function for limits

### LINKED ISSUE #1151: multer dynamic limit size
hi.
 I wanna have a dynamic limit size which is stored in memory cache but I cant.

my multer middleware :
```js
const multer = require("multer");
const SettingModel = require("../../models/Setting");
const SequelizeService = require("../app/sequelize/sequelize.service");

    cb(null, "./uploads/");
  },
    cb(null, file.originalname.replace(/\s+/g, ""));
  },
});

const fileFilter = (req, file, cb) => {
  if (
    file.mimetype === "image/jpeg" ||
    file.mimetype === "image/png" ||
    file.mimetype === "image/webp"
  ) {
    cb(null, true);
  } else {
    cb(new Error("Not an image! Please upload an image.", 400), false);
  }
};

exports.upload = multer({
  storage: storage,
  limits: {
    fileSize: 5 * 1024 * 1024,
  },
  onError: function (err, next) {
    next(err);
  },
  fileFilter: fileFilter,
});

```
`
and its my router 

```js
const express = require("express");
let router = express.Router();
const UserController = require("./user.controller");
const upload = require("../../middleware/upload").upload;

router.post(
  "/create",
  upload.single("photo"),
  UserController.store
);

module.exports = router;

```
 when I change multer limit to : cache.get("limit") it wont be updated when i update cache

hossein-zare:
`limits` can now be a function that receives the request and returns the limits object, called once per request before parsing starts. The returned object is validated like a static one; errors go to `next(err)`. A static `limits` object behaves as before.

Documented in the README (limits are per request, not per file), changelog entry included.

Closes #1151

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- Accept a function for `limits`, called with the request, to set limits per request ([#1133](https://github.com/expressjs/multer/pull/1133))
 - Files skipped by `fileFilter` no longer count towards `maxCount` ([#1426](https://github.com/expressjs/multer/pull/1426))
 - Accept requests with exactly `limits.parts` parts; `LIMIT_PART_COUNT` now fires only when the limit is exceeded. If you set `parts` one higher to work around this, you can drop the extra one ([#1446](https://github.com/expressjs/multer/pull/1446))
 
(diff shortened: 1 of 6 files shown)


UlisesGascon (approved):
Thanks @hossein-zare for this contribution! I did update the PR against the current code base and extended the initial idea with a more solid approach that also include tests 👍

PR #1143, opened 2022-09-30
wrong syntax fixed in readme.md ln 159

ctafsiras:


dougwilson:
This follows the convertion to signal optional args from the Node.js docs. See https://nodejs.org/dist/latest-v18.x/docs/api/http.html#agentcreateconnectionoptions-callback for instance.

PR #1154, opened 2022-10-20
Add option to disable detection

osm:


LinusU:
Do you have any background on why you would want to do this?

PR #1156, opened 2022-11-07
[ReadME Example] - include file extension to save it properly

mcpeblocker:
Example was not working: files were being saved without their extension according to the example shown for DiskStorage


LinusU:
Sorry for the late reply.

This is being done intentionally since it could be a security vulnerability to blindly use the extension provided by the client.

The client is free to send whatever extension it wants, and web-servers (or even just browsing the files with a file manager) could treat the file differently depending on the extension.

Issue #1015, opened 2021-06-18
UTF-8 Problem


xcandev:
Hello,

This is my File Name
- Ferdi Özbeğen - Gündüzüm Seninle.mp3

This is multer originalname
- Ferdi O\bzbeg\u0006en - Gu\bndu\bzu\bm Seninle.mp3

Turkish characters are normally supported in UTF-8. How can I edit this?

HassanDev13:
same thing !

serhatleventyavas:
I have the same problem. How can I solve it?

PR #1158, opened 2022-11-19
accept utf8 filenames

almontasser:


almontasser:
This PR should fix #1015

jhpung:
This feature already opened 😅

Please check this PR => https://github.com/expressjs/multer/pull/1102

dumbasPL:
maybe It would be nice to be able to specify what encoding to use instead. Changing the default encoding will be a breaking change. What about people that don't want/can't use utf8?

LinusU:
Sorry for the late response on this.

It feels like this is a breaking change? In that case I don't see how we could get it into the 1.x release, but maybe we can do the same change in the 2.x branch? (#399)

dumbasPL:
> It feels like this is a breaking change?

yes, because currently, people are using the `Buffer.from(filename, 'latin1').toString('utf8');` trick. 
If we parse an utf8 string as latin1 it will spit out garbage.

That's why I suggested only adding an option to change the encoding as a config option but leaving the default one as is to not break existing setups.

> but maybe we can do the same change in the 2.x branch?

sounds good as well

Doc999tor:
@LinusU I created a different PR that doesn't break backward compatibility - it extends the multer's API with a default `defParamCharset: 'latin1'` value
Also, improved the translations
https://github.com/expressjs/multer/pull/1210

UlisesGascon:
Thanks @almontasser! This landed as an option instead, to keep the default backwards compatible: `multer({ defParamCharset: 'utf8' })` (#1210, available since 2.1.0). Closing this one.

PR #1161, opened 2022-11-22
feat: added cjs support via esbuild

titanism:
@LinusU we have added CJS support without breaking anything or making any drastic changes.  can you please merge and publish a new rc?  thank you! 🙏 🙇


bjohansebas:
I don't see a reason why the code needs to be compiled unless it's to generate the TypeScript files. Is there any reason why you can't use Multer in .cjs files?

PR #1165, opened 2022-12-01
dependancy-update [Security-Issue] [high]

RashadTanjim:
"busboy": "^1.6.0",
    "mocha": "^10.1.0",

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   ],
   "dependencies": {
     "append-field": "^1.0.0",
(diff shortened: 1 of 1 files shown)


LinusU:
An update of Busboy is already submitted as https://github.com/expressjs/multer/pull/1097.

Since this is a breaking change we cannot land it on the stable 1.x branch.

You can use version `1.4.5-lts.1` if you are on Node.js 10.x or newer, for older versions of Node.js there is currently no solution...

Issue #1104, opened 2022-06-09
Issue with UTF-8 characters in filename


CleyFaye:
Hi,

I found recently that something changed regarding the handling of filename containing utf-8 characters; they seem to be passed as-is, which was not the case before.

After investigating a bit I could reproduce the issue with the minimal code in https://github.com/CleyFaye/test-multer

I found that the browser side just pass the name as-is in the "filename" part of the header.
I've seen another issue related to using "filename*", but there is two problem with that: the browser's formdata does not use this, and [RFC7578](https://datatracker.ietf.org/doc/html/rfc7578#section-4.2) actually says it should not be used.

What would be the proper way to handle this? Obviously it is possible, server side, to convert the content of `originalname` by putting all characters as bytes in an array then interpreting it as an utf-8 string (it does work), but since I never had this issue with older versions, I suspect something changed in the way multer handles this.

CleyFaye:
The small test provided returned the expected filename with multer@1.4.4, and changed with multer@1.4.4-lts.1.

dvantage:
Same problem after update from 1.4.4 on 1.4.5-lts.1

dvantage:
Multer has nothing to do with it, Busboy has changed something.

This solved my problem:
<pre>
file.originalname = Buffer.from(file.originalname, 'latin1').toString('utf8')
</pre>

CleyFaye:
Multer has something to do about this, since it definitely changed behavior in an arguably incompatible way in what looks like a patch revision.

What to do however I'm not sure; either way would be fine (interpreting the utf-8 to be consistent with previous behavior or passing the raw string to not make assumptions about encoding), but I believe this kind of change in a patch is troublesome to users.

georgerdp-cic:
> Multer has nothing to do with it, Busboy has changed something. [mscdex/busboy#20](https://github.com/mscdex/busboy/issues/20)
> 
> This solved my problem:
> 
> file.originalname = Buffer.from(file.originalname, 'latin1').toString('utf8')

God bless you. 

I've managed to make a bode in my app 
``` 
    const fileName = Buffer.from(el.originalname, 'latin1').toString('utf8');
````

because in my case "invalid £$ file.txt" was becoming "invalid Â£$ file.txt".
 Ideally we have this fixed when busboy is fixing that end. 
Thanks a lot.

sominlee74:
[...]

Issue #1194, opened 2023-03-22
multipart/form-data windows-1251


UncleVic:
I have a problem with decoding messages in the charset `windows-1251`.
If a form contains charset=windows-1251, data can't be decoded correctly. After parsing a request a string will be corrupted.

My payload, for example. The tag `<ErrorText>` contains the charset `windows-1251`
```
POST /api/v1/events/providers/ipay HTTP/1.1
X-Forwarded-Proto: https
Connection: close
Content-Length: 673
Content-Type: multipart/form-data; charset=windows-1251; boundary=BS_20230321154848
User-Agent: BS_SOU_749

--BS_20230321154848
Content-Disposition: form-data; name="XML"

<?xml version="1.0" encoding="windows-1251" ?>
<ServiceProvider_Request>
  <DateTime>20230321154848</DateTime>
  <Version>1</Version>
  <RequestType>TransactionResult</RequestType>
  <ServiceNo>1</ServiceNo>
  <PersonalAccount>2gzD17a7MHFpnQGZmGag75</PersonalAccount>
  <Currency>933</Currency>
  <RequestId>6888</RequestId>
  <TransactionResult>
    <TransactionId>356624</TransactionId>
    <ServiceProvider_TrxId>2gzD17a7MHFpnQGZmGag75</ServiceProvider_TrxId>
    <CardTerminal>888888</CardTerminal>
    <ErrorText>Îïåðàöèÿ îòìåíåíà</ErrorText>
  </TransactionResult>
</ServiceProvider_Request>
--BS_20230321154848--
```

Doc999tor:
I opened a PR for fixing this issue - https://github.com/expressjs/multer/pull/1210
Waiting for a review
@UncleVic maybe you could use the fork I until the PR is merged

PR #1210, opened 2023-05-23
headers-support-utf8

Doc999tor:
This PR adds support of [busboy](https://github.com/mscdex/busboy/blob/d08cc9cc5572a29354480f6ec829a042c68d8bac/README.md#L145)'s `defParamCharset` option for supporting UTF8 characters in request headers such as `filename`
Default: `'latin1'` for backward compatibility
Not required

Resolves https://github.com/expressjs/multer/issues/1104, https://github.com/expressjs/multer/issues/1194

Example of usage: 
```javascript
multer({
	storage,
	defParamCharset: 'utf8',
})
```


Doc999tor:
TypeScript definitions update - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65578

starnayuta:
@Doc999tor
Thanks for the PR.
I am also having trouble with UTF-8 file names.


The default for defParamCharset should be UTF-8.
Here is the parsing of the filenames for each version that I recognize

1.4.4 and below: UTF-8
1.4.5-lts.1: Latin-1

If it takes semantic versioning, it should be UTF-8, the same as 1.4.4.


@LinusU wrote a comment that might be relevant. What do you think?
https://github.com/expressjs/multer/issues/1104#issuecomment-1500010854

At any rate, I hope it will be released as soon as possible.
Thank you.

Doc999tor:
@starnayuta 
I understand that but the v1.4.5 was marked as lts, so I don't believe it will be changed as part of 1.4.5; only as a new version, until it will become lts as well

Doc999tor:
@LinusU Are there any blockers for merging this PR? Can we merge it?

Doc999tor:
@UlisesGascon, maybe you could assist with this PR?

threshold862543:
Would be nice to merge this.

PRR24:
It has been over a year, please kindly merge it.

bjohansebas:
Hi @Doc999tor, you can add tests to verify the behavior

luiscabus:
> Hi @Doc999tor, you can add tests to verify the behavior

it just works, get another task to add tests for this, the current behavior isnt even well tested anyways

Darkhogg:
@bjohansebas This change is simply "pass this option down to busboy", so a test for it should basically just be "busbuy has received the argument", right?
@luiscabus If you need help getting that done I can lend you a hand, I'm a bit blocked by this PR and would love to see it merged and released.

bjohansebas:
Well, I really have no idea if it's actually possible to test this, but since it fixes something in the package, ideally, we should have a test for it.

bjohansebas:
Could you please do a rebase? And an example of tests could be:

```js

describe('UTF-8 filenames', function () {
  var uploadDir, upload

  beforeEach(function (done) {
    temp.mkdir(function (err, path) {
      if (err) return done(err)

      var storage = multer.diskStorage({
        destination: path,
          cb(null, file.originalname)
        }
      })

      uploadDir = path
      upload = multer({ storage: storage, defParamCharset: 'utf-8' })
      
      done()
    })
  })

  afterEach(function (done) {
    rimraf(uploadDir, done)
  })

  it('should handle UTF-8 filenames', function (done) {
    var req = new stream.PassThrough()
    var boundary = 'AaB03x'
    var body = [
      '--' + boundary,
      'Content-Disposition: form-data; name="small0"; filename="ÖoϪ.dat"',
      'Content-Type: text/plain',
      '',
      'test with UTF-8 filename',
      '--' + boundary + '--'
    ].join('\r\n')

    req.headers = {
      'content-type': 'multipart/form-data; boundary=' + boundary,
      'content-length': body.length
    }

    req.end(body)

    upload.single('small0')(req, null, function (err) {
      assert.ifError(err)

      assert.strictEqual(req.file.originalname, 'ÖoϪ.dat')
      assert.strictEqual(req.file.fieldname, 'small0')

      done()
    })
  })
})
```

Doc999tor:
I will do, thanks for the test example

ShubhamOulkar:
Just a remainder 

> ```js
> filename="ÖoϪ.dat"'
> ```

Non-ASCII characters (like Ö or Ϫ) can be misinterpreted if the stream isn’t properly encoded. To avoid ambiguity and ensure cross-platform compatibility, use the filename* syntax with UTF-8 percent-encoding as recommended in [RFC 5987, Section 3.2.2](https://datatracker.ietf.org/doc/html/rfc5987#section-3.2.2).

Also, see [this test case](https://github.com/expressjs/multer/blob/main/test/unicode.js#L34) in the Multer codebase before making changes to how filenames are handled.

planted-mreinstein:
Why is this not merged yet? I appreciate the desire for good test coverage but this is very clearly broken, and it's not a very complex change, just passing through a parameter to busboy.

Doc999tor:
@bjohansebas rebased and added tests, sorry for taking too long 🙏 

* should use latin1 as default charset for non-extended parameters
* should use custom charset when defParamCharset is specified
* should not affect extended parameters with explicit charset
* should handle different charsets correctly
* should work with memory storage
* should work with array uploads

unc0ded:
Interesting coincidence, I had just completed working on a new PR with your changes and a few tests 😆

index.js
```diff
@@ -19,6 +19,7 @@ function Multer (options) {
 
   this.limits = options.limits
   this.preservePath = options.preservePath
+  this.defParamCharset = options.defParamCharset || 'latin1'
```
unc0ded, line 22: This default shouldn't need to be passed explicitly, busboy uses latin1 anyway when the defParamCharset passed is undefined.

Doc999tor: Yes, I know - I added it in case something will change in `busboy` default settings, `multer` will still behave consistently 
@LinusU @bjohansebas @UlisesGascon what do you think? I can change it if it's unnecessary precaution

Doc999tor:
@UlisesGascon @unc0ded @bjohansebas @LinusU Can one of you review the PR? 
@unc0ded Is my comment ok from your perspective?

unc0ded:
> @UlisesGascon @unc0ded @bjohansebas @LinusU Can one of you review the PR? @unc0ded Is my comment ok from your perspective?

I think busboy won't change the default without a major version bump anyway, so setting a default here may be overkill. But I leave it upto the maintainers here to take a call on that.

ducfilan:
It should be merged!

LowBoBYT:
It should be merged!

PR #1211, opened 2023-05-25
Firebase function: Add check if rawBody exist in request 

lamberto15:
In the latest version of Firebase Functions, the req.body is already processed by Firebase Functions, which is why Multer is not compatible anymore. However, by adding the following code, Multer will be compatible again with Firebase Functions:

`if (req.rawBody) {
` 
`    busboy.end(req.rawBody)`
`} else {`
 `   req.pipe(busboy);`
`}`
This code checks if req.rawBody exists and if it does, it ends the busboy instance with req.rawBody. If req.rawBody is not available, it pipes the req object to busboy for further processing.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -9,12 +9,12 @@ var MulterError = require('./multer-error')
 var FileAppender = require('./file-appender')
 var removeUploadedFiles = require('./remove-uploaded-files')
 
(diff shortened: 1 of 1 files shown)


RobbieProkop:
Where specifically are you adding this if else?

UlisesGascon:
Superseded by #1466

PR #1216, opened 2023-06-12
Fix file exists but file name does not exist

Relsoul:
[https://www.rfc-editor.org/rfc/rfc1867](https://www.rfc-editor.org/rfc/rfc1867)
file name is an optional field.

eg.
```
----------------------------052948472259119770453321
Content-Disposition: form-data; name="filepond"
Content-Type: application/octet-stream
```

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -96,7 +96,15 @@ function makeMiddleware (setup) {
     // handle files
     busboy.on('file', function (fieldname, fileStream, filename, encoding, mimetype) {
       // don't attach to the files object, if there is no file
-      if (!filename) return fileStream.resume()
+      if (!filename) {
+        // @link https://github.com/mscdex/busboy/blob/master/lib/types/multipart.js
+        // If the type is 'application/octet-stream', it is also identified as a file 
+        if(mimetype==='application/octet-stream'){
+          filename = 'unknownFile'
+        }else{
+          return fileStream.resume()
+        }
+      }
 
       // Work around bug in Busboy (https://github.com/mscdex/busboy/issues/6)
       if (limits && Object.prototype.hasOwnProperty.call(limits, 'fieldNameSize')) {


lib/make-middleware.js
```diff
@@ -96,7 +96,15 @@ function makeMiddleware (setup) {
...
-      if (!filename) return fileStream.resume()
+      if (!filename) {
+        // @link https://github.com/mscdex/busboy/blob/master/lib/types/multipart.js
+        // If the type is 'application/octet-stream', it is also identified as a file 
+        if(mimetype==='application/octet-stream'){
+          filename = 'unknownFile'
+        }else{
+          return fileStream.resume()
+        }
+      }
```
ljwagerfield, line 107: Hey team! I've been thinking about the current implementation of this `if` statement. It might be beneficial for us to consider allowing `undefined` filenames. This change would give developers more flexibility in handling cases where a client does or does not provide a `filename` field. Such scenarios can be pretty important for certain applications.

I wanted to share an interesting point that @Relsoul brought up regarding RFC 7578. The RFC mentions that the `filename` in the content disposition is recommended but not mandatory. It states:

> For form data that represents the content of a file, a name for the file SHOULD be supplied as well, by using a 'filename' parameter of the Content-Disposition header field.

With this in mind, it seems quite likely that developers using Multer would appreciate the ability to handle RFC-compliant uploads in a way that best suits their needs, especially when a filename isn't provided. For example, in our system, we prefer storing `NULL` in the database for the "original file name" if the uploader doesn't provide it, instead of assigning a default value.

Interestingly, `busboy` allows for undefined filenames, which is a great feature for flexibility. I think it might be a good idea for multer to adopt a similar approach. What do you all think?

To gain parity with busboy, it would simply be a case of removing this entire `if` statement, and acknowledging that the `originalname` field may be `undefined` to reflect RFC 7578:

```suggestion
```

Relsoul:
yep,
Here is a simple demo: [https://github.com/Relsoul/upload-demo](https://github.com/Relsoul/upload-demo)

In the browser environment, formData.append(name, value, filename);

The filename is optional, but the browser will automatically add a filename for Blob/File type objects. However, in other environments, the filename may not be automatically added. When I use a buffer for direct upload, I might ignore the filename.

In Node Express, this is reflected in that if the filename does not exist, it skips the destination function and filename function.

UlisesGascon:
Thanks @Relsoul! Files without a `filename` are a known gap in v2 (#455). Accepting them, with `originalname` left undefined rather than inventing a name, is already in the 3.0.0 alpha. v2 stays as-is since changing which parts land in `req.files` would break existing apps.

Issue #1240, opened 2024-02-08
Update Docs: Error: read ECONNRESET thrown during test using supertest, mocha but works fine while using postman. 


sebakthapa:
## ECONNRESET during test only
```js
    Error while testing invalid file upload Error: read ECONNRESET
        at TCP.onStreamRead (node:internal/stream_base_commons:217:20) {
      errno: -4077,
      code: 'ECONNRESET',
      syscall: 'read',
      response: undefined
    }
```

## What is causing the error?
One of the reason why this error occurs during test but not on other methods is related to  how  we call `cb()` of fileFilter option/function when we are rejecting certain file.

The above error was thrown when `fileFilter` is passed and inside it the callback is returned as:
 ```js
 cb(new Error("some error msg");
```

but not when callback is returned as:
   ```js
 cb(null, false);
```

## What is missing in the documentation?
The right way to pass `cb()` for custom error is the following:

```js
cb(null, false, new Error("some error msg"))
```

## What should be added in the documentation and where?
```js
// If you want to reject a file with custom error:
cb(null, false, new Error("some error msg"))
```

The above line should be appended in the [fileFilter option description](https://github.com/expressjs/multer?tab=readme-ov-file#filefilter)

It would be better to append it below the line which shows how to reject a file.

sebakthapa:
This contains misinformation as there is not third parameter for the `cb()` function.

PR #1241, opened 2024-02-08
Updated readme

sebakthapa:
### Added 
Added reject with custom error line of code in [fileFilter option description's code block](https://github.com/expressjs/multer?tab=readme-ov-file#filefilter) 

This points to issue #1240

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,11 +290,14 @@ function fileFilter (req, file, cb) {
   // The function should call `cb` with a boolean
   // to indicate if the file should be accepted
 
+  // To accept the file pass `true`, like so:
+  cb(null, true)
+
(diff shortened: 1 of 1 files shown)


LinusU:
I can't see that `fileFilter` is looking at the third argument being passed?

https://github.com/expressjs/multer/blob/aa42bea6ac7d0cb8fcb279b15a7278cda805dc63/lib/make-middleware.js#L115

sebakthapa:
Yes, you are right. 
I'm extremely sorry, for any inconvenience that you faced due to my silly mistake.

I am also concerned if we can reject files with a custom error. Using the first parameter of the `cb()` returns `ECONNRESET` on test but works fine on req made with Postman.

Issue #1233, opened 2023-10-04
TypeError: Cannot read properties of undefined (reading 'length') when fieldNameSize is not set


dvtradeling:
We use Nest.js FilesInterceptor which uses multer under the hood.

When fieldname is not provided, but fieldNameSize is set, the above line is still executed and TypeError is thrown accordingly.
Check this - https://github.com/expressjs/multer/blob/master/lib/make-middleware.js#L103

We need to prevent it.

<img width="905" alt="Screenshot 2023-10-04 at 12 39 44" src="https://github.com/expressjs/multer/assets/92851293/249d3468-8441-405b-a583-94d2d776678c">

For fixing this issue please add this line
<img width="1745" alt="Screenshot 2023-10-04 at 12 53 18" src="https://github.com/expressjs/multer/assets/92851293/2651b004-dce5-43c8-a7f3-1378f6f57332">

joeyguerra:
Can you set the fieldname to an empty string?

Markiz9999:
Is there a workaround to prevent this vulnerability?

joeyguerra:
Have you tried passing a field name?

Markiz9999:
Sure. With the passed field name everything works fine. But the important thing here is to prevent the application from crashing when the client does something illegal (in this case, send a request with an empty field name).

If your question is did I try to send an empty string in the field name, then by default it is an empty string, but in that case the busboy library emits an undefined value.

https://github.com/mscdex/busboy/blob/master/lib/types/multipart.js#L296
```javascript
partName = undefined;
```

https://github.com/mscdex/busboy/blob/master/lib/types/multipart.js#L313
```javascript
if (disp.params.name)
  partName = disp.params.name;
```

https://github.com/mscdex/busboy/blob/master/lib/types/multipart.js#L358
```
this.emit(
  'file',
  partName,
  this._fileStream,
  { filename,
    encoding: partEncoding,
    mimeType: partType }
);
```

joeyguerra:
I see. Fix #913 was released in [1.44](https://github.com/expressjs/multer/pull/913#issuecomment-987846653). Perhaps upgrading to at least 1.44 would address this issue.

ctcpip:
@ImLunaHey the tl;dr is: governance and security overhaul within the express org. for more info please see the links below.  thanks!

- https://expressjs.com/2025/01/09/rewind-2024-triumphs-and-2025-vision.html
[...]

PR #1256, opened 2024-04-23
Fixes expressjs/multer#1233. Makes multer handle missing field names.

Markiz9999:
Without this fix fields without a name result in a "TypeError: Cannot read property 'length' of undefined".

The current change allows getting an error from multer that makes it possible to handle it in servers.

See: https://github.com/expressjs/multer/issues/1233


ctcpip:
released in 2.0.1. thank you!

Issue #1218, opened 2023-06-30
Multer & Express.js File Upload request hanged (never ending pending)


agussetyar:
I'm working with very simple file upload feature using Express.js & Multer.

This is my multer middleware code:
```javascript
const path = require('path')
const multer = require('multer')

const storage = multer.diskStorage({
  destination: function (req, file, cb) {
    // Set the destination folder for uploaded files
    cb(null, './public/uploads/')
  },
  filename: function (req, file, cb) {
    // Generate a unique filename to avoid naming conflicts
    const uniqueSuffix = Date.now() + '-' + Math.round(Math.random() * 1e9)
    const originalFileExtension = path.extname(file.originalname)
    const filename = file.fieldname + '-' + uniqueSuffix + originalFileExtension

    cb(null, filename)
  },
})

const upload = multer({
  storage: storage,
  limits: { fileSize: 100 * 1024 * 1024 },
  fileFilter: function (req, file, cb) {
    const allowedTypes = /pdf|xls|zip|rar|jpg/
    const fileType = path.extname(file.originalname).toLowerCase()

    if (allowedTypes.test(fileType) === false) {
      const err = new Error('File extensions is not allowed.')
      err.status = 500
      err.name = 'FileValidationError'

      cb(err, false)
    }

    cb(null, true)
  },
})

module.exports = upload
```

This is my error handler middleware:
```javascript
// wrap unathorized error
router.use((err, req, res, next) => {
  if (['UnauthorizedError', 'FileValidationError'].includes(err.name)) {
    return res.status(err.status).json({
      error: true,
      code: err.status,
      message: err.message,
    })
  }

  next(err)
})
```

When I hit this endpoint from Frontend (React.js) using Axios and FormData. I got never ending 'pending' request and I dont know why this is happen with this simple code. The expected results is custom JSON response should be returned.
![pending](https://github.com/expressjs/multer/assets/30733517/47d4da80-554b-49da-98cd-7719b9df832f)

After a minutes searching, I found that if I remove the 'FileValidationError' from error handler logic or using this code:
```javascript
if (err instanceof multer.MulterError) {}
```
The request is not pending and it returned response the Error() instance.
![catchError](https://github.com/expressjs/multer/assets/30733517/c27df5c1-981e-45dd-88a2-73c168b081d4)

But, this behavior only happens in the Frontend side. I tested with Postman and the request looks fine.

Can someone tell me If I do something wrong in this situation?

baselshoban:
[...]

PR #1267, opened 2024-08-07
Fixes expressjs/multer#1218. No response in Pending status.

bettep-dev:
When a file upload is in progress in the native app, a network communication timeout occurs in the Pending state, 

but I verified that the busboy.on('close') event does not fall into the Pending state.

I modified the code to handle errors in storage._handleFile.

See: [#1218](https://github.com/expressjs/multer/issues/1218)

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -113,18 +113,13 @@ function makeMiddleware (setup) {
       var placeholder = appender.insertPlaceholder(file)
 
       fileFilter(req, file, function (err, includeFile) {
-        if (err) {
-          appender.removePlaceholder(placeholder)
-          return abortWithError(err)
-        }
-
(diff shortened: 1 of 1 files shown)


pwliuab:
this works,

yfhui:
thx

thingasd:
bro strong :sorry:

UlisesGascon:
Thanks @bettep-dev! The hang from #1218 no longer happens on 2.x: when `fileFilter` rejects a file, multer drains the rest of the request before calling `next(err)`, and a large rejected upload gets its 400 right away (verified on 2.3.0). This change would also turn `cb(null, false)` (skip this file and continue) into an abort of the whole request, which is documented behaviour we need to keep.

PR #1294, opened 2025-03-20
test: add test for out-of-band error event

LinusU:
Cheery picks the test from the `lts` branch to `master`


mhassan1:
@LinusU We may want to backport an improvement to this test introduced in https://github.com/expressjs/multer/pull/1296.

Issue #1176, opened 2023-01-14
Uncaught `error` event from busboy can crash http servers


max-mathieu:
There is a case of malformed requests that can take down a full nodejs app, due to an uncaught `error` event thrown by `busboy`

Using the latest `1.4.5-lts.1` from npm, the following code results in a full crash of the app (tested on both node 14 and node 18)

```
const express = require('express')
const multer  = require('multer')
const http  = require('http')
const upload = multer({ dest: 'uploads/' })
const port = 8888

const app = express()

app.post('/upload', upload.single('file'), function (req, res) {
  res.send({})
})

app.listen(port, () => {
  console.log(`Listening on port ${port}`)

  const boundary = 'AaB03x'
  const body = [
    '--' + boundary,
    'Content-Disposition: form-data; name="file"; filename="test.txt"',
    'Content-Type: text/plain',
    '',
    'test without end boundary'
  ].join('\r\n')
  const options = {
    hostname: 'localhost',
    port,
    path: '/upload',
    method: 'POST',
    headers: {
      'content-type': 'multipart/form-data; boundary=' + boundary,
      'content-length': body.length,
    }
  }
  const req = http.request(options, (res) => {
    console.log(res.statusCode)
  })
  req.on('error', (err) => {
    console.error(err)
  })
  req.write(body)
  req.end()
})
```

Result:
```
# node multer-bug.js
Listening on port 8888
events.js:377
      throw er; // Unhandled 'error' event
      ^

Error: Unexpected end of form
    at Multipart._final (/Users/max/src/tests/node_modules/busboy/lib/types/multipart.js:588:17)
    at callFinal (internal/streams/writable.js:610:10)
    at processTicksAndRejections (internal/process/task_queues.js:82:21)
Emitted 'error' event on Multipart instance at:
    at emitErrorNT (internal/streams/destroy.js:106:8)
    at emitErrorCloseNT (internal/streams/destroy.js:74:3)
    at processTicksAndRejections (internal/process/task_queues.js:82:21) {
  storageErrors: []
}
# back to prompt
```

The crash doesn't happen as soon as `process.on('uncaughtException')` is set, which probably explains why this test passes https://github.com/expressjs/multer/blob/lts/test/error-handling.js#L226-L250 with the same request.

This is not happening with multer 1.4.3 (and busboy pre-1.0)

[...]

PR #1177, opened 2023-01-14
Fix out-of-band error event from busboy

max-mathieu:
Fixes https://github.com/expressjs/multer/issues/1176

PR #1296, opened 2025-04-10
fix: handle two busboy error events

mhassan1:
This PR enables `multer` to handle two out-of-band error events from `busboy`; it does this by replacing `process.nextTick` (which happens too soon after the first error event) with `setImmediate`.

I don't know if this fix will handle more than two, but I also don't think `busboy` will ever emit more than two.

This is a follow-up to https://github.com/expressjs/multer/pull/1177.

Resolves https://github.com/expressjs/multer/issues/1176.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -41,7 +41,7 @@ function makeMiddleware (setup) {
       if (isDone) return
       isDone = true
       req.unpipe(busboy)
-      process.nextTick(() => {
+      setImmediate(() => {
         busboy.removeAllListeners()
       })
       next(err)
--- a/test/express-integration.js
+++ b/test/express-integration.js
@@ -150,4 +150,50 @@ describe('Express Integration', function () {
     req.write(body)
     req.end()
   })
+
+  it('should not crash on malformed request that causes two errors to be emitted by busboy', function (done) {
+    var upload = multer()
+
+    app.post('/upload2', upload.single('file'), function (req, res) {
+      res.status(500).end('Request should not be processed')
+    })
+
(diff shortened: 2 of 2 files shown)


mhassan1:
@LinusU Please take a look, when you have a chance. Thanks.

mhassan1:
@UlisesGascon @LinusU Please take a look, when you have a chance. Thanks.

LinusU:
Could you duplicate the test into a new one, instead of modifying the current one, so that we are sure that the previous single error is also handled? 🙏

mhassan1:
> Could you duplicate the test into a new one, instead of modifying the current one, so that we are sure that the previous single error is also handled? 🙏

@LinusU Done, please take a look.

bjohansebas (approved):

mhassan1:
@UlisesGascon @LinusU Please take a look, when you have a chance. Thanks.

UlisesGascon (approved):

UlisesGascon:
Not sure why the [CI pipeline](https://github.com/expressjs/multer/blob/master/.github/workflows/ci.yml) was not triggered :thinking:

bjohansebas:
@UlisesGascon see #1302

ctcpip:
this looks good, but we are pulling in this fix as part of another PR.  I will update when that gets merged.  thank you!

ctcpip (changes requested):
waiting to merge, please hold

ctcpip:
resolved via https://github.com/expressjs/multer/commit/2c8505f207d923dd8de13a9f93a4563e59933665

thanks everyone!

PR #1303, opened 2025-05-11
support node >=18

bjohansebas:
As with other packages, we support Node >=18 🙂 for this new version

### DIFF
--- a/.github/workflows/legacy.yml
+++ b/.github/workflows/legacy.yml
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        node-version: [12, 13, 14, 15, 16, 17]
+        node-version: [14, 15, 16, 17]
         # Node.js release schedule: https://nodejs.org/en/about/releases/
 
     name: Node.js ${{ matrix.node-version }} - ${{matrix.os}}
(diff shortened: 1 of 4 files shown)


ctcpip:
> As with other packages, we support Node >=18 🙂 for this new version

this is not really the case.  see [non-goals in the ADR](https://github.com/expressjs/discussions/blob/master/docs/LTS-strategy.md#non-goals)

therefore, we should not drop support for older versions if we don't have a reason beyond "it's old node"

ctcpip (changes requested):
see comment above

UlisesGascon (commented):
We can check if there is a need for engines change once all the dependencies are updated

bjohansebas:
@ctcpip We need to update type-is, which supports Node >= 18 :)

.github/workflows/legacy.yml
```diff
@@ -0,0 +1,96 @@
+name: legacy
+
+on:
+  push:
+    branches:
+      - explore-new-api
```
ctcpip, line 6: renamed

UlisesGascon (approved):
LGTM! Please don't forget to update the `CHANGELOG.md` file as this is a breaking change :)

UlisesGascon (approved):

PR #1304, opened 2025-05-11
convert package to commonjs

bjohansebas:
The package is being switched back to CommonJS since the new API is currently in ESM. For more information, see https://github.com/expressjs/discussions/pull/323

### DIFF
--- a/lib/read-body.js
+++ b/lib/read-body.js
@@ -1,14 +1,13 @@
-import { extname } from 'node:path'
-import { pipeline as _pipeline } from 'node:stream'
-import { promisify } from 'node:util'
+const { extname } = require('node:path')
+const { pipeline: _pipeline } = require('node:stream')
+const { promisify } = require('node:util')
 
-import Busboy from '@fastify/busboy'
-import { createWriteStream } from 'fs-temp'
-import hasOwnProperty from 'has-own-property'
-import _onFinished from 'on-finished'
-import FileType from 'stream-file-type'
+const Busboy = require('@fastify/busboy')
+const { createWriteStream } = require('fs-temp')
+const _onFinished = require('on-finished')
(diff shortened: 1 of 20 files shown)


LinusU:
@bjohansebas the reason I switched to ESM initially was in order to consume [the `file-type` package](https://www.npmjs.com/package/file-type) (via `stream-file-type`)

Here is some more thoughts on it: https://github.com/expressjs/multer/issues/1042#issuecomment-973902270

Sorry for having been so slow to respond to things! Unfortunately I have had a very hectic time lately, and I no longer use Multer professionally so can't spend company time on it 😞 

If you need any updates to `stream-file-type` I'd be happy to assist (could also lift it into appropriate org), or anything else you need here! Thank you for spending time on this 🙏

lib/read-body.js
```diff
@@ -1,14 +1,13 @@
...
 
-import Busboy from '@fastify/busboy'
-import { createWriteStream } from 'fs-temp'
-import hasOwnProperty from 'has-own-property'
-import _onFinished from 'on-finished'
-import FileType from 'stream-file-type'
+const Busboy = require('@fastify/busboy')
+const { createWriteStream } = require('fs-temp')
+const _onFinished = require('on-finished')
+const FileType = require('stream-file-type')
```
LinusU, line 8: (I believe that this will fail on Node.js versions that cannot `require` ESM modules)

bjohansebas:
Hey, thanks @LinusU for the historical reason, i hadn’t found it before.

Given that it’s really complicated to do, the options I see are:

- Stick with ESM, we’d need to request an exception for this [ADR](https://github.com/expressjs/discussions/pull/323)
- Move support to Node >=22.12, which would go against the LTS plan

I think the best option will be to stay with ESM.

PR #1325, opened 2025-05-23
📝 list languages in table to prevent GH right-aligning list due to RTL language

ctcpip:
note: I did not update this in the translated readmes because they are all in different states/age and most do not list all the languages, so would require additional translation.  probably safe to just do a machine translation, but that could be done as a follow-up


ctcpip:
### before/afer:

Screenshot 2025-05-23 at 11 04 37 AM

PR #1326, opened 2025-05-26
ci: add windows

### LINKED ISSUE #1035: No works on windows
I try use this lib on windows10 and not work :(, on my another laptop with ubuntu 20.04 if work

bjohansebas:
fixed https://github.com/expressjs/multer/issues/1035

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "devDependencies": {
     "deep-equal": "^2.0.3",
     "express": "^4.13.1",
-    "form-data": "^1.0.0-rc1",
+    "form-data": "^4.0.2",
     "fs-temp": "^1.1.2",
     "mocha": "^11.3.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
   ],
   "dependencies": {
     "append-field": "^1.0.0",
-    "busboy": "^1.0.0",
+    "@fastify/busboy": "^1.0.0",
```
ShubhamOulkar, line 23: Test are failed because of changes in busboy

bjohansebas: No, actually that change shouldn't be here. I was testing some things and accidentally committed it here

ShubhamOulkar: 👍

UlisesGascon:
Windows support in CI was added in mutler@2.3.0

PR #1327, opened 2025-05-26
feat: add charset option for multipart forms

### LINKED ISSUE #1104: Issue with UTF-8 characters in filename

bjohansebas:
Rebase of https://github.com/expressjs/multer/pull/1210, and I’m removing the changes to the READMEs since I’d prefer a native speaker to handle them. Although, I still hope to eventually remove the README translations from here.

closes #1104
closes https://github.com/expressjs/multer/pull/1210

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ Key | Description
 `fileFilter` | Function to control which files are accepted
 `limits` | Limits of the uploaded data
 `preservePath` | Keep the full path of files instead of just the base name
+`charset` | For multipart forms, the default character set to use for values of part header parameters (e.g. filename) that are not extended parameters (that contain an explicit charset). **Default**: `'latin1'`.
 
 In an average web app, only `dest` might be required, and configured as shown in
 the following example.
--- a/index.js
+++ b/index.js
@@ -19,6 +19,8 @@ function Multer (options) {
 
   this.limits = options.limits
   this.preservePath = options.preservePath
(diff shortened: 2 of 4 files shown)


ShubhamOulkar (changes requested):

UlisesGascon:
we're keeping busboy's option names as they are in v2. `defParamCharset` has been public since 2.1.0, and #1465 adds `defCharset`, `highWaterMark` and `fileHwm` under the same rule, so renaming one of them now would mean an alias and a deprecation for the rest of the v2 line for a cosmetic gain. If we want friendlier names, that's a v3 decision for all of them at once.

PR #1334, opened 2025-06-06
Test: Enhance cross-platform test reliability and update CI workflow

ShubhamOulkar:
This PR updates test assertions for cross platforms. All 61 tests were run and passed successfully on both Windows and Linux environments. Also made changes in CI workflow.

Tasks

- [X] update test files for cross platforms
- [X] update CI workflow 

Issue taken from https://github.com/expressjs/multer/issues/1035#issuecomment-2933748349

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
           npm-i: mocha@8.4.0
           
         - name: Node.js 11.x
-          node-version: "11.0.0"
+          node-version: "11.15"
           npm-i: mocha@8.4.0
           
         - name: Node.js 12.x
@@ -62,7 +62,7 @@ jobs:
           npm-i: mocha@9.2.2
           
         - name: Node.js 13.x
-          node-version: "13.0.0"
+          node-version: "13.14"
           npm-i: mocha@9.2.2
           
         - name: Node.js 14.x
@@ -89,16 +89,16 @@ jobs:
           node-version: "20.0.0"
           
         - name: Node.js 21.x
-          node-version: "21.0.0"
+          node-version: "21"
           
         - name: Node.js 22.x
-          node-version: "22.0.0"
+          node-version: "22"
 
         - name: Node.js 23.x
-          node-version: "23.0.0"
+          node-version: "23"
 
         - name: Node.js 24.x
-          node-version: "24.0.0"
+          node-version: "24"
 
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,55 +50,55 @@ jobs:
 
         include:
         - name: Node.js 10.x
-          node-version: "10"
+          node-version: "10.16.0"
           npm-i: mocha@8.4.0
           
         - name: Node.js 11.x
-          node-version: "11"
+          node-version: "11.0.0"
           npm-i: mocha@8.4.0
           
         - name: Node.js 12.x
-          node-version: "12"
+          node-version: "12.0.0"
           npm-i: mocha@9.2.2
           
         - name: Node.js 13.x
-          node-version: "13"
+          node-version: "13.0.0"
           npm-i: mocha@9.2.2
           
         - name: Node.js 14.x
-          node-version: "14"
+          node-version: "14.0.0"
           npm-i: mocha@9.2.2
           
         - name: Node.js 15.x
-          node-version: "15"
+          node-version: "15.0.0"
           npm-i: mocha@9.2.2
           
         - name: Node.js 16.x
-          node-version: "16"
+          node-version: "16.0.0"
           
         - name: Node.js 17.x
-          node-version: "17"
+          node-version: "17.0.0"
           
         - name: Node.js 18.x
-          node-version: "18"
+          node-version: "18.0.0"
           
         - name: Node.js 19.x
-          node-version: "19"
+          node-version: "19.0.0"
           
         - name: Node.js 20.x
-          node-version: "20"
+          node-version: "20.0.0"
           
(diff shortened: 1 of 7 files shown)


UlisesGascon (approved):

bjohansebas (changes requested):
Hey, if you could add these changes to the workflow to verify that Windows really works well in the tests. https://github.com/expressjs/multer/pull/1326/files

ShubhamOulkar:
>   should handle unicode filenames:
     Uncaught AssertionError [ERR_ASSERTION]: ifError got unwanted exception: ENOENT: no such file or directory, open
      at C:\a\multer\multer\test\unicode.js:54:14

Should we ignore this test on Windows? Or change the file name to some other Unicode character that is supported by windows.

ShubhamOulkar (commented):
LGTM

.github/workflows/ci.yml
```diff
@@ -1,21 +1,38 @@
...
     runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+
+      - name: Install dependencies
+        run: npm install --ignore-scripts --include=dev
```
wojtekmaj, line 27: `npm ci` would have been a better choice but we're missing a lockfile 🫣

.github/workflows/ci.yml
```diff
@@ -1,21 +1,38 @@
...
+
+      - name: Run lint
+        run: npm run lint
```
wojtekmaj, line 30: In latest lts we can already use `node --run`

.github/workflows/ci.yml
```diff
@@ -55,23 +72,23 @@ jobs:
           npm-i: mocha@9.2.2
           
         - name: Node.js 15.x
-          node-version: "15.14"
+          node-version: "15.14.0"
```
wojtekmaj, line 75: That made me wonder, should we really be testing against the latest versions of old Node.js majors? What if we break compat with 15.0.0 by accident?

ShubhamOulkar:
```suggestion
          node-version: "15"
```
yup, this should a nice change.

.github/workflows/ci.yml
```diff
@@ -1,21 +1,38 @@
 name: ci
 
 on:
-- pull_request
-- push
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '*.md'
```
bjohansebas, line 8:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -1,21 +1,38 @@
...
+  pull_request:
+    paths-ignore:
+      - '*.md'
```
bjohansebas, line 11: This prevents us from having to bypass the branch rules when they are enabled.
```suggestion
```

.github/workflows/ci.yml
LinusU: > `npm ci` would have been a better choice but we're missing a lockfile 🫣

Is it good to check in a lock file in a library though? That lockfile won't be respected when publishing to npm, so we wouldn't be testing with the same dependencies as used by our users, but instead whatever dependencies happened to be used when we generate the lockfile.

test/disk-storage.js
```diff
@@ -3,14 +3,21 @@
...
 var path = require('path')
 var util = require('./_util')
 var multer = require('../')
 var temp = require('fs-temp')
 var rimraf = require('rimraf')
 var FormData = require('form-data')
 
+function assertFileProperties (file, name) {
+  const expectedSize = util.fileSizeByName(name)
+  assert.strictEqual(file.fieldname, path.parse(name).name)
```
LinusU, line 15: We should probably have tests where the fieldname and name of the file doesn't match. Otherwise this wouldn't catch if we in the library somehow mix them together 🤔

.github/workflows/ci.yml
wojtekmaj: One could argue that adding a lockfile is a double-edged sword, but personally I think not having a lockfile in place creates far more problems than it solves. 

Solves:
1. You will have your code tested against the latest (matching) version of declared dependencies.
    Counterargument: one can set up dependabot or renovate to update dependencies automatically or at least semi-automatically.
2. No lockfile is less code, less code is less maintenance and fewer occasions for potential conflicts.

Creates:
1. Let's say one of your dependencies release a new (matching) version. Unfortunately it introduced a change that causes your tests to go red. Suddenly:
- Your tests on main branch are failing - even though they were passing yesterday and you didn't make any changes, or
- Contributors are raising PRs that unexpectedly go red, causing confusion: did I break anything?
2. Your installs are significantly slower on CI as the entire dependency tree must be resolved from scratch.
3. People technically working on the same branch may have different lockfiles locally, causing your code to be non-deterministic: a common root cause of a classic "weird, it works on _my_ machine!"

.github/workflows/ci.yml
```diff
@@ -35,43 +50,43 @@ jobs:
 
         include:
         - name: Node.js 10.x
-          node-version: "10.24"
+          node-version: "10"
```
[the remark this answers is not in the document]
ShubhamOulkar: I think this will fail because of dev dependencies compatibility. I prefer `latest` but let's test it 🤔.

LinusU (approved):
Great work! 🙌

UlisesGascon (approved):
Rebased on main and squashed to two commits (tests + ci). Kept your changes; adapted to #1353/#1355/#1452 which landed meanwhile, pinned the actions per our current policy, and ran coverage only on the Ubuntu leg. Windows on Node < 22 is marked continue-on-error for now... I'll make it blocking in a follow-up

PR #1335, opened 2025-06-11
fix: validate all limits at construction time

### LINKED ISSUE #1132: `MulterError: File too large` not thrown when `fileSize` is floating point
When uploading files larger than 1MB with `limits: { fileSize: 1024 }`  I get the expected `MulterError: File too large` error but when using `limits: { fileSize: 1024.1 }` the file upload succeeds which should not.

**Expectation**: Multer should be able to handle unexpected config inputs.

ShubhamOulkar:
busboy compares most limits (`files`, `fields`, `parts`, `fileSize`, `fieldSize`) with strict equality, so a non-integer value never matches and silently disables that limit. #1395 fixed this for `fileSize`; this extends the same check to every key in `limits`: each must be a non-negative integer or `Infinity` (the documented default), otherwise `multer()` throws a `TypeError`. No new error code; config errors stay `TypeError`s like the existing `options` check.

Rebased onto `main` as a single commit with the original approach simplified (no `MulterError` for construction-time errors, `Infinity` accepted).

### DIFF
--- a/lib/multer-error.js
+++ b/lib/multer-error.js
@@ -8,7 +8,8 @@ var errorMessages = {
   LIMIT_FIELD_VALUE: 'Field value too long',
   LIMIT_FIELD_COUNT: 'Too many fields',
   LIMIT_UNEXPECTED_FILE: 'Unexpected field',
-  MISSING_FIELD_NAME: 'Field name missing'
+  MISSING_FIELD_NAME: 'Field name missing',
+  LIMIT_OPTION_ERROR: 'Limit option must be an integer'
 }
 
 function MulterError (code, field) {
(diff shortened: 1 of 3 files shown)


lib/multer-error.js
```diff
@@ -8,15 +8,19 @@ var errorMessages = {
...
 function MulterError (code, field) {
   Error.captureStackTrace(this, this.constructor)
   this.name = this.constructor.name
   this.message = errorMessages[code]
   this.code = code
-  if (field) this.field = field
+  if (field) {
+    this.field = field
+    if (code === 'LIMIT_OPTION_ERROR') this.message += `: ${field}`
+  }
```
efekrskl: I'm not sure if we should give `LIMIT_OPTION_ERROR` special treatment. `field` is already a part of the context, so imo we shouldn't mutate the message

UlisesGascon (approved):
I made some tweaks and included all the changes from `main` to fit into the existing code base. Thanks @ShubhamOulkar for the hard work!

PR #1344, opened 2025-07-17
deps: update dependencies to latest versions

wusfe:
-  Update the version of  `type-is`

multer@2.0.1
`-- type-is@1.6.18
  `-- @0.3.0

In this `makeMiddleware ` function, we used the `media-typer` module  provided by `type-is`.
However, the lower versions of `media-typer` do not have the `test` method

Version 0.3.0 of `media-typer ` does not provide the ` test` method, which will result in an error in the production environment，

- Update the version of  `express`

Avoid the impact of `type-is` version desynchronization

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,12 +24,12 @@
     "concat-stream": "^2.0.0",
     "mkdirp": "^0.5.6",
     "object-assign": "^4.1.1",
-    "type-is": "^1.6.18",
+    "type-is": "^2.0.1",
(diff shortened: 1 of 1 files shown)


UlisesGascon:
Thank you for your PR, @wusfe! We truly appreciate your contribution. While we won't be able to accept this change at the moment since the dependency upgrades would break compatibility with Node@10. Looking forward to more contributions from you :smile:

PR #1350, opened 2025-08-08
chore: drop mkdirp dependency

wojtekmaj:
fs.mkdir gained `recursive` option in Node.js v10.12.0 (source: https://nodejs.org/api/fs.html#fsmkdirpath-options-callback). Minimum version of Node.js supported by multer at the moment is v10.16.0, so we should be good to go.


UlisesGascon (approved):
Great one! Thanks @wojtekmaj :clap:

PR #1351, opened 2025-08-08
chore: drop object-assign dependency

wojtekmaj:
`Object.assign` is natively supported since Node.js v4.0.0. Minimum version of Node.js supported by multer at the moment is v10.16.0, so we should be good to go.


UlisesGascon (approved):
Great work @wojtekmaj!

PR #1352, opened 2025-08-08
chore: drop xtend dependency

wojtekmaj:
Spread in object literals is natively supported in Node.js since v8.3.0. Minimum version of Node.js supported by multer at the moment is v10.16.0, so we should be good to go.


LinusU (approved):
Neat! 👍

PR #1356, opened 2025-08-15
refactor: remove concat-stream dependency

Phillip9587:
Removes `concat-stream` from the dependencies. busboy file streams only emit Buffers, so `MemoryStorage` collects the chunks and `Buffer.concat`s them itself; the one test that used it does the same. This drops `concat-stream` and its transitive dependencies (`readable-stream`, `buffer-from`, `typedarray`, `inherits`) from the production install.

Rebased onto `main` as a single commit with the minimal implementation (no generic buffer coercion, prototype style like the rest of `storage/`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,10 +25,10 @@
   "dependencies": {
     "append-field": "^1.0.0",
     "busboy": "^1.6.0",
-    "concat-stream": "^2.0.0",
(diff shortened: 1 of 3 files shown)


UlisesGascon (approved):
I did made some tweaks and rebased the changes. Great work @Phillip9587!

PR #1368, opened 2026-02-16
Change default `defParamCharset` from `latin1` to `utf-8`

UlisesGascon:
BREAKING CHANGE: The default character set for parsing multipart form-data headers (e.g., filenames) is now 'utf-8' instead of 'latin1'.

This aligns with modern web standards where UTF-8 is ubiquitous. Users who need the previous behavior can explicitly set: `multer({ defParamCharset: 'latin1' })`

Continuation of https://github.com/expressjs/multer/pull/1210


bjohansebas (changes requested):
We already had the `v3` branch, right? Those changes should go there. But more importantly, I don’t think that option is even in `v3` yet, which would be worth double-checking.

PR #1382, opened 2026-03-18
test: accept files exactly at fileSize limit

### LINKED ISSUE #1348: multer throws "File too large" error if file size of exact "limits.fileSize" bytes provided.
Multer throws "File too large" error if file size of exat "`limits.fileSize`" bytes provided. It is non-intuitive behavior. (Or it might be a corner-case.)

It seems that [busboy](https://github.com/mscdex/busboy) emits event "`limit`" when limits.fileSize bytes is hit. However, this means that busboy will not further process files, and it *might be* truncated.

https://github.com/expressjs/multer/blob/2e07e77311b966b9ae5cbd63cb3d6c34501d291e/lib/make-middleware.js#L153-L156

Relevant lines in busboy: https://github.com/mscdex/busboy/blob/6b3dcf69d38c1a8d53a0b3e4c88ba296f6c91525/lib/types/multipart.js#L476-L482

### Current Workaround 

increase `limits.fileSize` by 1.

## Environment information

**Version**:

```js
"dependencies": {
    // ...
    "express": "^5.1.0",
    "multer": "^2.0.2",
    // ...
}
```

**Platform**:
`Linux NX-281041 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux`

**Node.js version**:
`v22.16.0`

abhu85:
## Summary

Fixes #1348

Busboy emits the `'limit'` event when `fileSize` equals `limits.fileSize`, causing multer to reject files that are exactly at the configured limit. This is an off-by-one error - files **at** the limit should be accepted; only files **exceeding** the limit should be rejected.

### Root Cause

In [busboy's multipart.js](https://github.com/mscdex/busboy/blob/6b3dcf69d38c1a8d53a0b3e4c88ba296f6c91525/lib/types/multipart.js#L476-L482):

```js
if (fileSize === fileSizeLimit) {
  this._fileStream.emit('limit');
  // ...
}
```

The `'limit'` event fires when the file size equals the limit, not when it exceeds it. This is arguably correct behavior for busboy (it can't know if more bytes are coming), but multer should handle this case gracefully.

### Solution

Pass `fileSize + 1` to busboy so that files at the exact limit don't trigger the limit event. Files 1 byte over the configured limit will still trigger the event correctly.

```js
var busboyLimits = limits
if (limits && typeof limits.fileSize === 'number') {
  busboyLimits = Object.assign({}, limits, { fileSize: limits.fileSize + 1 })
}
```

## Test Plan

- [x] Added test: file exactly at limit is accepted
- [x] Added test: file 1 byte over limit is rejected  
- [x] Added test: empty file accepted when limit is 0
- [x] Added test: non-empty file rejected when limit is 0
- [x] All existing tests pass (76 total)

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -26,6 +26,15 @@ function makeMiddleware (setup) {
     var preservePath = options.preservePath
     var defParamCharset = options.defParamCharset
 
+    // Work around off-by-one issue in busboy: it emits 'limit' event when
+    // fileSize === limits.fileSize, even though the file should be accepted.
+    // We pass fileSize + 1 to busboy and enforce the exact limit ourselves.
+    // See: https://github.com/expressjs/multer/issues/1348
+    var busboyLimits = limits
+    if (limits && typeof limits.fileSize === 'number') {
+      busboyLimits = Object.assign({}, limits, { fileSize: limits.fileSize + 1 })
+    }
+
     req.body = Object.create(null)
 
     var busboy
@@ -126,7 +135,7 @@ function makeMiddleware (setup) {
     try {
       busboy = Busboy({
         headers: req.headers,
-        limits: limits,
+        limits: busboyLimits,
         preservePath: preservePath,
         defParamCharset: defParamCharset
       })
--- a/test/error-handling.js
+++ b/test/error-handling.js
@@ -66,6 +66,70 @@ describe('Error Handling', function () {
     })
   })
 
+  it('should accept file exactly at fileSize limit', function (done) {
+    // tiny0.dat is 122 bytes - set limit to exactly that
+    var form = new FormData()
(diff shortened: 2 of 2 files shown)


abhu85:
Hi maintainers 👋

Friendly ping on this PR fixing the fileSize boundary off-by-one issue (#1348). The fix ensures files exactly at the limit are accepted rather than rejected.

All 76 tests pass. Would appreciate a review when you get a chance. Thanks!

UlisesGascon (approved):

UlisesGascon (approved):
Thanks @abhu85! I solved the conflicts and only the tests will be merged at the end. Thanks for your contribution to the project ❤️

Issue #1389, opened 2026-04-05
Teardown timing in make-middleware causes uncaught Busboy errors and multi-error regression


mukeshdhadhariya:
### Problem

Multer’s teardown logic in `make-middleware.js` has a timing issue when handling Busboy errors.

There are two related behaviors:

1. In malformed multipart requests, Busboy may emit a late `error` after teardown begins.
2. In some cases (e.g. Express integration tests), Busboy emits multiple errors.

### Current Behavior

With `process.nextTick` used for teardown cleanup:

- A second Busboy error can surface as:
  - `Unexpected end of form`
  - `done() called multiple times`
- This is reproducible using the existing Express integration test.

### Reproduction

Run:

npx mocha --reporter spec --exit --check-leaks express-integration.js --grep "two errors to be emitted by busboy"

### Observed Output

- Uncaught Error: Unexpected end of form
- done() called multiple times

### Root Cause

Teardown scheduling affects ordering between:
- Busboy error emission
- Multer cleanup
- Express completion flow

When multiple errors are emitted, current timing allows a second error to escape or trigger duplicate completion.

### Notes

This indicates that teardown timing alone (`process.nextTick`) is not sufficient to safely handle all Busboy error scenarios.

A fix needs to ensure:
- Late Busboy errors are safely handled
- Multiple error emissions do not trigger duplicate completion

jonchurch:
Your repro is "run the tests"? your report does not repro, even after fixing your repro command:
`npx mocha --reporter spec --exit --check-leaks test/express-integration.js --grep "two errors to be emitted by busboy"`

are you trying to revert adfeaf6? Please explain

PR #1390, opened 2026-04-05
Fix Busboy teardown race and multi-error handling during multipart parsing

mukeshdhadhariya:
### Summary

This PR addresses a teardown timing issue #1389  in `make-middleware.js` that affects Busboy error handling.

Two related problems are observed:

1. Busboy may emit a late `error` after teardown begins.
2. In malformed multipart flows, Busboy can emit multiple errors, causing:
   - uncaught `Unexpected end of form`
   - `done()` being called multiple times

### Root Cause

Listener cleanup timing interacts with Busboy's error emission order.  
Using `process.nextTick` alone does not fully guarantee safe handling when multiple errors are emitted.

### Evidence

Failing test:

npx mocha --reporter spec --exit --check-leaks express-integration.js --grep "two errors to be emitted by busboy"

Observed failures:

- Uncaught Error: Unexpected end of form
- Error: done() called multiple times

### Fix

This change ensures teardown and error handling remain consistent by preventing duplicate completion and ensuring late Busboy errors are safely absorbed.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -48,7 +48,7 @@ function makeMiddleware (setup) {
       isDone = true
       if (busboy) {
         req.unpipe(busboy)
-        setImmediate(() => {
+        process.nextTick(() => {
           busboy.removeAllListeners()
         })
       }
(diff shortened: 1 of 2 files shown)


jonchurch:
slop

PR #1391, opened 2026-04-15
feat(disk-storage): support opt-in flush for durable uploads (#1381)

### LINKED ISSUE #1381: uploaded files are not flushed to disk
It looks like multer doesn't call `sync` or `datasync` when files are closed, which can lead to data corruption.

https://github.com/expressjs/multer/blob/368c8a10cca11854cf17c24029fefd1eafb1c059/storage/disk.js#L37

NodeJS does support a `flush` option which will ensure data is flushed to disk before closing the file. It would be great if multer could add an option to set this.

https://nodejs.org/api/fs.html#filehandlecreatewritestreamoptions

## Environment information

**Version**:

v2.1.1

**Platform**:

Linux

**Node.js version**:

v25.8.1

Upload a file, and observe that in certain conditions the handler may say that the upload has completed, but in reality the data has not been written to disk and so the file is either corrupt or lost.

SAY-5:
Closes #1381.

## Why

As reported in #1381, the disk storage pipes the uploaded file into a vanilla \`fs.createWriteStream\` and reports the upload as successful the moment the stream finishes. On Linux (and any OS with a write-back page cache) the bytes can still be sitting in kernel buffers when the request handler returns, a power loss or crash at the wrong moment leaves the file truncated or zero-length, but the HTTP response already claimed success.

Node.js 20.10 / 21.0 added a \`flush\` option to \`fs.createWriteStream\` that calls \`fdatasync()\` on the file descriptor before the stream closes, which is exactly what this workflow needs.

## Change

Expose \`flush\` as an opt-in \`DiskStorage\` option:

\`\`\`js
multer.diskStorage({
  destination: './uploads',
  flush: true
})
\`\`\`

- Default stays \`false\`, so existing users get the exact same buffered-write behavior and the same throughput, no perf regression.
- When \`flush\` is \`true\`, the option is forwarded as \`fs.createWriteStream(path, { flush: true })\`. On runtimes older than Node 20.10, \`createWriteStream\` silently ignores the option, which matches the documented upstream behavior.
- No change to \`_removeFile\` or error handling.

## Tests

Added two regression tests in \`test/disk-storage.js\`:

1. With \`flush: true\`, the test wraps \`fs.createWriteStream\` and asserts it was called with \`{ flush: true }\` while the upload still round-trips at the expected size.
2. Without \`flush\`, \`createWriteStream\` is still called with no options, verifies the default behavior is unchanged.

Full \`npm test\` suite: **74/74 passing**.

### DIFF
--- a/storage/disk.js
+++ b/storage/disk.js
@@ -15,6 +15,14 @@ function getDestination (req, file, cb) {
 
 function DiskStorage (opts) {
   this.getFilename = (opts.filename || getFilename)
+  // When `flush` is truthy, the disk storage asks the underlying write
+  // stream to fdatasync() the uploaded file before `_handleFile` reports
+  // success, so callers that expect the data to survive a crash or power
+  // loss (see #1381) can rely on it. The option is forwarded as
+  // `fs.createWriteStream(..., { flush: true })`, which was added in
+  // Node.js 20.10 / 21.0. On older runtimes the option is ignored and
+  // behavior falls back to the previous buffered-write semantics.
+  this.flush = opts.flush === true
 
   if (typeof opts.destination === 'string') {
     fs.mkdirSync(opts.destination, { recursive: true })
@@ -34,7 +42,8 @@ DiskStorage.prototype._handleFile = function _handleFile (req, file, cb) {
       if (err) return cb(err)
 
       var finalPath = path.join(destination, filename)
-      var outStream = fs.createWriteStream(finalPath)
+      var writeStreamOptions = that.flush ? { flush: true } : undefined
+      var outStream = fs.createWriteStream(finalPath, writeStreamOptions)
 
       file.stream.pipe(outStream)
       outStream.on('error', cb)
--- a/test/disk-storage.js
+++ b/test/disk-storage.js
@@ -185,4 +185,59 @@ describe('Disk Storage', function () {
       done()
     })
   })
+
+  it('should support the flush option for durable uploads (#1381)', function (done) {
+    var calls = []
+    var realCreateWriteStream = fs.createWriteStream
+    fs.createWriteStream = function (filePath, options) {
+      calls.push({ filePath: filePath, options: options })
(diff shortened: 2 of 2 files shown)


MohammedAlkindi (commented):
Ran this branch on Windows 11 (Node v24.18.0, so `{ flush: true }` is actually exercised — it needs Node ≥ 20.10) on a default `core.autocrlf=true` clone, since CI has no Windows lane and the new tests assert exact on-disk byte sizes.

**As-is:**

```
64 passing, 13 failing
```

Before that number alarms anyone: **12 of the 13 are not this PR's fault.** They are the pre-existing fixture problem — no `.gitattributes`, so a Windows clone smudges the `.dat` fixtures to CRLF (`small0.dat`: 1778 → 1803 bytes on disk) and every hardcoded-size assertion in the existing suite fails the same way on the base too.

The 13th, though, is this PR's own new test, and it fails for the same reason it will fail for every Windows contributor:

```
should support the flush option for durable uploads (#1381):
AssertionError: 1803 !== 1778     ← assert.strictEqual(req.file.size, 1778)
```

**The flush mechanism itself is fine on NTFS.** As a diagnostic I replaced only the two hardcoded `1778`s in the new test with sizes measured from the fixture and re-ran: both new tests pass — `createWriteStream` receives `{ flush: true }`, the upload completes, and the on-disk size matches what was sent. So the durability path works here; it's the size idiom that doesn't travel.

Suggestion: assert against a measured fixture size rather than a literal (that's the direction #1406 takes for the existing tests, and it makes the new tests robust regardless of clone settings). Disclosure: my own open #1450 fixes the same underlying smudge by adding `.gitattributes` for the fixtures, so I'm not neutral on that corner — but the measured-size form is worth having even with attributes in place.

Also noting for the queue: the branch is currently conflicting with `main` (9 commits behind, including `9c801c7`, which touched `storage/disk.js` nearby), so it needs a rebase regardless.

PR #1394, opened 2026-04-23
fix(docs): handle string destination in storage engine template

### LINKED ISSUE #1278: "Potential Issue with destination Parameter Type in Multer Custom Storage"
in this file https://github.com/expressjs/multer/blob/master/StorageEngine.md
its giving a template for making a custom storage engine but there is a problem I guess..

function MyCustomStorage (opts) {
  this.getDestination = (opts.destination || getDestination)
}
function MyCustomStorage (opts) {
  this.getDestination = (opts.destination || getDestination)
}

MyCustomStorage.prototype._handleFile = function _handleFile (req, file, cb) {
  this.getDestination(req, file, function (err, path) {
    if (err) return cb(err)

    var outStream = fs.createWriteStream(path)

    file.stream.pipe(outStream)
    outStream.on('error', cb)
    outStream.on('finish', function () {
      cb(null, {
        path: path,
        size: outStream.bytesWritten
      })
    })
  })
}
its calling this.getDestination as function but it may not be a function if user provided "destination property" in the constructor..
or I am getting this wrong?

Bojun-Vvibe:
Closes #1278

## Repo
expressjs/multer

## Issue
https://github.com/expressjs/multer/issues/1278

## Root cause
The `StorageEngine.md` custom storage template assigned `opts.destination`
directly to `this.getDestination` and then invoked it as a function in
`_handleFile`. If a user follows the convention used by `DiskStorage` and
passes `destination` as a string path, the template throws a TypeError
because a string is not callable.

## Fix
Updated the `MyCustomStorage` constructor in `StorageEngine.md` to mirror
the behavior of `storage/disk.js`: when `opts.destination` is a string,
wrap it in a `function ($0, $1, cb) { cb(null, opts.destination) }`
adapter; otherwise fall back to the function-or-default assignment.

## Regression test
None. The change is documentation-only (`StorageEngine.md`); the existing
`storage/disk.js` already correctly handles string destinations and is
covered by `test/disk-storage.js`.

## Risk
trivial

## Verification
skipped: documentation-only change with no executable code path; existing
disk-storage tests cover the equivalent runtime behavior the docs now
reflect.

### DIFF
--- a/StorageEngine.md
+++ b/StorageEngine.md
@@ -44,7 +44,11 @@ function getDestination (req, file, cb) {
 }
 
 function MyCustomStorage (opts) {
-  this.getDestination = (opts.destination || getDestination)
+  if (typeof opts.destination === 'string') {
+    this.getDestination = function ($0, $1, cb) { cb(null, opts.destination) }
+  } else {
+    this.getDestination = (opts.destination || getDestination)
+  }
 }
 
 MyCustomStorage.prototype._handleFile = function _handleFile (req, file, cb) {


UlisesGascon:
Hey @Bojun-Vvibe! Can you explain a bit the motivation for this change and if you are working on it?

PR #1399, opened 2026-05-11
fix: preserve existing request body fields

### LINKED ISSUE #1187: [BUG] Multer overriding req.body contents, if any of express middleware add custom keys to req.body
Hello there, In the project i was working, I have integrated multer and found the strange bug wqhere multer overrides req.body data if any of middleware add data to req.body.
For Eg:
Before calling multer, there are couple of middlewares like ratelimiter, authentication verification. These controllers add some meta data to req.body and this is used in the final controller. When multer is called in one of middleware, it overrides all req.body data which was attached by previous middlewares and assigns the multer form data 's json to req.body. 

As a result, previous middleware's important data is lost.  Unlike multer, `express-file-upload` does the things correctly as observed.

JSON returned by multer,
```
 [Object: null prototype] {
  clientid: 'WFeNx12Fju800004',
  real_filename: 'uploadedcode.png',
  unix: '1676026107',
}
```

The same json returned by `express-file-upload`
```
{
  clientid: 'WFeNx12Fju800004',
  real_filename: 'uploadedcode.png',
  unix: '1676026107',
  limit: 66,
  auth: true,
  acl: "owner:read:write"
}
```
From above jsons the keys `limit, auth, acl` are overridden in case of multer and are lost, but using `express-file-upload` the keys are preserved.

raashish1601:
## Summary

Preserve existing own `req.body` properties when Multer initializes multipart parsing, instead of replacing them with an empty object.

Multer still uses a null-prototype object for `req.body`; it copies existing own enumerable properties into that object before appending multipart fields.

Fixes #1187.

## Validation

- `npm run lint`
- `npx mocha --reporter spec --exit --check-leaks test/fields.js`

I also ran `npm test`; it reached the new regression successfully, but the full suite fails in this Windows checkout on existing fixture byte-size and POSIX path assertions unrelated to this change, for example `1803 !== 1778` in storage/file size tests and `/testforme-` path matching.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -26,7 +26,15 @@ function makeMiddleware (setup) {
     var preservePath = options.preservePath
     var defParamCharset = options.defParamCharset
 
+    var existingBody = req.body
+
     req.body = Object.create(null)
+
+    if (existingBody && typeof existingBody === 'object') {
+      Object.keys(existingBody).forEach(function (key) {
+        req.body[key] = existingBody[key]
+      })
+    }
 
     var busboy
     var appender = null
--- a/test/fields.js
+++ b/test/fields.js
@@ -34,6 +34,37 @@ describe('Fields', function () {
     })
   })
 
+  it('should preserve existing body properties', function (done) {
+    var form = new FormData()
+
+    form.append('name', 'Multer')
+
+    form.getLength(function (err, length) {
+      if (err) return done(err)
+
+      var req = new stream.PassThrough()
+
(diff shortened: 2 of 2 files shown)


UlisesGascon:
Thanks @raashish1601! I am closing this for the same reasons as #907 and #445: `req.body` is the output of multer's own parsing, and copying an existing body in lets client-sent fields merge into values set by earlier middleware. Data that doesn't come from the request body belongs on another property (e.g. `req.user`), or that middleware can run after multer.

PR #1400, opened 2026-05-11
Fix multipart boundaries containing equals signs

### LINKED ISSUE #1051: Multer skips the request, when boundary contains an equal sign "="
https://github.com/expressjs/multer/blob/dd3ecef8dcac79568b96313dca00adad7563b45d/lib/make-middleware.js#L18

**type-is** use media-typer which treats the equal sign "=" as a separator (RFC 2616 sec 3.7). However, in the RFC 2046 Page 21 the boundary value is allows to use the equal sign "=".

So, when the boundary contains the equal sign "=", Multer will always skip the request.

raashish1601:
Fixes #1051.

Multer used `type-is` to detect multipart requests before handing the request to Busboy. `type-is` rejects unquoted multipart boundary parameters containing `=`, even though those boundaries can be parsed by the multipart stream itself. The result was that Multer skipped the request entirely.

This changes the early multipart check to avoid parsing parameters and only inspect the media type while still honoring `type-is.hasBody(req)`. Before constructing Busboy, it normalizes only unquoted boundary values containing `=` by quoting the boundary in the headers passed to Busboy. The original request headers are not mutated.

Validation:
- `npx mocha --reporter spec --exit --check-leaks test/fields.js`
- `npm run lint`
- `npm test` was also run; it still fails on existing Windows fixture-size/path assertions unrelated to this change, while the new fields regression passed.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -13,9 +13,40 @@ function drainStream (stream) {
   })
 }
 
+function isMultipart (req) {
+  var type = req.headers && req.headers['content-type']
+
+  return type && is.hasBody(req) &&
+    type.split(';')[0].trim().toLowerCase().indexOf('multipart/') === 0
+}
+
+function normalizeMultipartHeaders (headers) {
+  var type = headers['content-type']
+
+  if (typeof type !== 'string' || type.indexOf('=') === -1) {
+    return headers
+  }
+
+  var normalized = type.replace(/(^|;\s*)boundary=([^";\s][^;]*)/i, function (match, prefix, boundary) {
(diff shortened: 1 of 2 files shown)


UlisesGascon:
Thanks @raashish1601! A boundary containing `=` is fine per RFC 2046, but HTTP requires such a parameter value to be quoted, and both `type-is` and busboy itself reject the unquoted form (busboy throws "Malformed content type" on it). Rewriting request headers in multer to accommodate a non-conforming client isn't something we want on the parsing path. The fix belongs on the client (quote the boundary); if that isn't possible, a small middleware before multer can quote it in `req.headers['content-type']`

PR #1401, opened 2026-05-12
fix: remove partial disk file on aborted uploads

### LINKED ISSUE #1217: File delete issue
When user is uploading a file, multer stores it on the disk. But if user aborts the upload, the "abort" event is caught but the file cannot be deleted until server is closed.

raashish1601:
## Summary

When a multipart upload is aborted before disk storage finishes, Multer can have an in-flight file that has not yet reached the `uploadedFiles` cleanup list. This tracks those in-flight files and removes them through the same cleanup path when the request aborts.

The disk storage engine now exposes destination, filename, and path metadata before piping starts so partial files can be unlinked if the request aborts mid-write.

Fixes #1217.

## Validation

- `npx mocha --reporter spec --exit --check-leaks test/error-handling.js`
- `npm run lint`

Note: I also tried `npm test` locally on Windows, but the full suite has fixture byte-size failures from CRLF line-ending conversion in existing test fixtures. The focused error-handling suite and lint both pass.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -35,6 +35,7 @@ function makeMiddleware (setup) {
     var errorOccured = false
     var pendingWrites = new Counter()
     var uploadedFiles = []
+    var filesInProgress = []
 
     function done (err) {
       var called = false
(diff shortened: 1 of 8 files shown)


raashish1601:
Superseded by compliant single PR branch naming; closing this one.

PR #1405, opened 2026-05-29
Update outdated links to Express error handling guide

krzysdz:
The links without language in URL are currently broken (https://github.com/expressjs/expressjs.com/issues/2343). Wherever a translated version is available the links point to it.


krzysdz:
Update: links without language in URL work again, but require 2 redirects. First 301 replacing `.html` with `/` and changing protocol to HTTPS, then meta refresh to the `/en/` version.

UlisesGascon (approved):

PR #1407, opened 2026-05-30
fix: allow files at exact fileSize limit

### LINKED ISSUE #1348: multer throws "File too large" error if file size of exact "limits.fileSize" bytes provided.

raashish1601:
## What changed
- Increase configured ileSize limit passed to Busboy by 1 byte only when limits.fileSize is explicitly set.
- This avoids treating an upload exactly equal to limits.fileSize as over-size while preserving the current error for larger uploads.
- Add regression test for exact-boundary acceptance in 	est/error-handling.js.

Fixes #1348 (off-by-one file size limit behavior).

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -20,6 +20,20 @@ function makeMiddleware (setup) {
     var options = setup()
 
     var limits = options.limits
+    var busboyLimits = limits
+
+    if (limits && Object.prototype.hasOwnProperty.call(limits, 'fileSize')) {
+      busboyLimits = {}
+      var key
+      for (key in limits) {
+        busboyLimits[key] = limits[key]
+      }
+
+      if (typeof limits.fileSize === 'number' && isFinite(limits.fileSize)) {
+        busboyLimits.fileSize = limits.fileSize + 1
+      }
+    }
+
     var storage = options.storage
     var fileFilter = options.fileFilter
     var fileStrategy = options.fileStrategy
@@ -126,7 +140,7 @@ function makeMiddleware (setup) {
     try {
       busboy = Busboy({
         headers: req.headers,
-        limits: limits,
+        limits: busboyLimits,
         preservePath: preservePath,
         defParamCharset: defParamCharset
       })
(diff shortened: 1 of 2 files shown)


MohammedAlkindi:
Confirmed independently, including the case your test doesn't cover.

On current `main` a 1500-byte file against `fileSize: 1500` is rejected, so the bug is real. With this merged into `main` (the branch is 9 behind, merges clean) 1500 is accepted and 1501 still fails `LIMIT_FILE_SIZE`, so the boundary moves by exactly one and no further. That over-by-one case is the one worth pinning, since `+1` could plausibly have shifted the error instead of removing it.

`fileSize: 0` accepts an empty file and rejects one byte, `Infinity` passes through the `isFinite` guard untouched, and `files: 1` still raises `LIMIT_FILE_COUNT`.

Suite went 79 to 80 passing; the single failure is a pre-existing Windows path assertion that fails on `main` too.

UlisesGascon (approved):

PR #1411, opened 2026-06-08
fix(v3): port CVE-2025-47944 / CVE-2025-47935

UlisesGascon:
Backports the malformed-body DoS fix from 2.x (commit 2c8505f) to the v3 branch.

The v3 rewrite introduced a different failure shape than 2.x: when `@fastify/busboy` is fed a malformed multipart body, it emits no events at all (no `file`, no `finish`, no `error`), leaving the middleware awaiting `Promise.all([fields, files, guard])` indefinitely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## unreleased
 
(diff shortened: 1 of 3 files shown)


UlisesGascon:
I am not sure why the CI was failing before... just re run make it pass again :thinking:... anyway the PR is ready for review

PR #1414, opened 2026-06-11
docs: clarify preservePath behavior

vibhor-aggr:
Clarifies what `preservePath: true` does for uploaded file metadata.

The README currently says it keeps the full path instead of the base name, but it does not explain where that path appears or that it is client-supplied input. This adds wording that:

- identifies `file.originalname` as the field affected by `preservePath`;
- notes that Multer does not create directories or sanitize the path;
- warns applications to normalize or validate `file.originalname` before using it in a custom filename or storage engine.

Refs #846.

Verification:
- `npm test` (72 passing)
- `npm run lint` with HOME pointed at the task workspace because the sandbox home is read-only
- `git diff --check HEAD~1..HEAD`

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ Each file contains the following information:
 Key | Description | Note
 --- | --- | ---
 `fieldname` | Field name specified in the form |
-`originalname` | Name of the file on the user's computer |
+`originalname` | Name of the file on the user's computer, or the full path when `preservePath: true` |
 `encoding` | Encoding type of the file |
 `mimetype` | Mime type of the file |
 `size` | Size of the file in bytes |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -155,6 +155,13 @@ the following example.
 const upload = multer({ dest: 'uploads/' })
 ```
 
+When `preservePath` is enabled, Multer passes the incoming filename through with
+any path segments provided by the client. This is exposed as `file.originalname`;
+it does not change the destination folder, create directories, or sanitize the
+path for you. Treat `file.originalname` as untrusted input if you use it in a
+custom `filename` or storage engine, and normalize or validate it before writing
+to disk.
```
UlisesGascon:
```suggestion
path for you. `file.originalname` is always client-supplied and should be treated
as untrusted; with `preservePath` it additionally contains the path segments the
client sent. Normalize or validate it before using it in a custom `filename` or
storage engine.
```

UlisesGascon (approved):
Thanks @vibhor-aggr! I just added a minor suggestion 👍

PR #1415, opened 2026-06-13
fix: add .gitattributes to prevent CRLF corruption of test fixtures on Windows

UjjwalKumar239:
## Problem
On Windows, Git's automatic CRLF line ending conversion corrupts binary 
test fixture files (.dat), causing 12 test failures due to file size 
mismatches (expected 1778 bytes, got 1803 bytes — a difference of 25 bytes 
matching exactly 25 `\r` characters added by Windows).

## Fix
- Added `.gitattributes` marking `test/files/*.dat` and `test/files/*.jpg` 
  as binary to prevent line ending conversion on any OS
- Fixed hardcoded Unix path separator `/testforme-` → `testforme-` in 
  `test/functionality.js` for cross-platform compatibility

## Testing
All 72 tests now pass on Windows.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+# Auto detect text files and perform LF normalization
+* text=auto
+
+# Treat test fixture files as binary (prevent CRLF conversion on Windows)
+test/files/*.dat binary
+test/files/*.jpg binary
--- a/test/functionality.js
+++ b/test/functionality.js
@@ -129,8 +129,8 @@ describe('Functionality', function () {
     util.submitForm(parser, form, function (err, req) {
       assert.ifError(err)
       assert.strictEqual(req.files.length, 2)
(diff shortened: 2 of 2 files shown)


MohammedAlkindi:
Both changes here landed through later PRs, so nothing in this one still applies: the `.gitattributes` via #1450 (opened 2026-08-10, mine) and the separator via #1452 (2026-08-16), both merged 2026-08-24.

This PR proposed both first, on 2026-06-13, and had no review in the 72 days before that.

The versions on main are wider. `test/files/** binary` also covers nested fixtures that `*.dat` misses, and `path.sep + 'testforme-'` rejects a directory named `NOTtestforme-`, which a bare `'testforme-'` accepts.

It no longer merges cleanly: `.gitattributes` add/add, plus a conflict in `test/functionality.js`.

PR #1416, opened 2026-06-13
feat: add filename to file errors

### LINKED ISSUE #607: Filesize in mutliple upload returns fieldname instead of filename

UjjwalKumar239:
Exposes the client-supplied file name as `err.filename` on `LIMIT_FILE_SIZE` (both the sync and the async-`fileFilter` path) and `LIMIT_UNEXPECTED_FILE` errors, next to the existing `err.field`, so an app can tell the user which file was rejected. Field-level errors are unchanged. Documented in the README as untrusted input, like `file.originalname`.

Rebased onto `main` as a single commit; the accidental reformatting of `make-middleware.js` was dropped.

Closes #607


UlisesGascon (approved):
LGTM! I just pushed the original changes on top of the current code base 👍

PR #1426, opened 2026-07-02
fix: do not consume maxCount for files skipped by fileFilter

### LINKED ISSUE #1419: fileFilter-skipped files consume same-field maxCount
## Summary

For `.array(name, maxCount)` and `.fields(...)`, Multer decrements the per-field `maxCount` before the user `fileFilter` decides whether the file should be accepted. A same-field file skipped with `cb(null, false)` can therefore consume the quota and cause the first acceptable file to be rejected with `LIMIT_UNEXPECTED_FILE`.

## Code path

Validator / consumer anchors:

- `index.js:31-33`: public field specs initialize per-field quota in `filesLeft`.
- `index.js:40-44`: quota is checked and decremented before the user filter runs.
- `index.js:45`: the user `fileFilter` runs after the decrement.
- `lib/make-middleware.js:204-206`: `includeFile === false` removes the placeholder and drains the stream, so the file is not accepted into `req.file(s)`.
- `lib/file-appender.js:56-57`: accepted files are later serialized into request file state.

## Steps to reproduce

Validation level: dynamic control-flow reproduction using the real `index.js` wrapper plus source-control-flow validation.

```js
const multer = require('multer')

const upload = multer({
  fileFilter: (_req, file, cb) => cb(null, file.originalname === 'keep.txt'),
}).array('docs', 1)

// Same request order:
// 1. docs=skip.txt  -> fileFilter returns false
// 2. docs=keep.txt  -> fileFilter would return true
```

Observed behavior from a local control-flow repro of the real wrapper:

```json
[
  {"err":null,"include":false},
  {"err":{"code":"LIMIT_UNEXPECTED_FILE","field":"docs"}}
]
```

## Expected behavior

`maxCount` should limit accepted files for the field. A file explicitly skipped by `fileFilter` should not consume the accepted-file quota, or the quota should be restored when `includeFile` is false.

## Actual behavior

The first skipped same-field file consumes the quota before inclusion is known. The next same-field file is rejected before the user `fileFilter` can accept it, even though zero files have been accepted.

## Existing coverage

I searched current issues and PRs for `fileFilter maxCount LIMIT_UNEXPECTED_FILE`, `fileFilter array maxCount`, `skipped file maxCount`, and broader `fileFilter` / `maxCount` terms. I found adjacent discussions such as #1057, #950, and #1093, but they do not appear to cover skipped files consuming same-field quota before any file is accepted.

## Suggested fix

Move the quota decrement until after `fileFilter` returns `includeFile === true`, or restore the quota when `includeFile` is false. Keep the current `LIMIT_UNEXPECTED_FILE` behavior for truly unexpected fields and accepted files beyond `maxCount`.

## Suggested tests

- Regression test for `.array('docs', 1)` where `docs=skip.txt` is filtered out and `docs=keep.txt` is accepted in the same request.
- Equivalent regression test for `.fields([{ name: 'docs', maxCount: 1 }])`.

---
Submitted with Codex.

Sagargupta16:
## Description

`wrappedFileFilter` decrements the field's remaining-count **before** the user's `fileFilter` runs:

```js
filesLeft[file.fieldname] -= 1
fileFilter(req, file, cb)
```

So a file the user skips with `cb(null, false)` still consumes a slot. With `.array('docs', 1)`, uploading a skipped file followed by an accepted file on the same field makes the second file fail with `LIMIT_UNEXPECTED_FILE` — even though **zero** files were stored. Per the README, `cb(null, false)` means the file is silently skipped (never added to `req.files`), so it must not count against `maxCount`. Fixes #1419.

## Fix

Move the decrement inside the `fileFilter` callback and only decrement when the file is actually accepted:

```js
fileFilter(req, file, function (err, includeFile) {
  if (!err && includeFile) filesLeft[file.fieldname] -= 1
  cb(err, includeFile)
})
```

This routes all callers (`.single` / `.array` / `.fields`) through one guard. Truly unexpected fields and accepted files beyond `maxCount` still error as before.

## Test

Added a regression test in `test/file-filter.js`: a `.array('docs', 1)` upload where the filter skips the first same-field file and accepts the second — asserts exactly one file (`tiny1.dat`) is stored with no error. The test fails on `main` without the fix (`LIMIT_UNEXPECTED_FILE`) and passes with it; existing fileFilter tests remain green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- Files skipped by `fileFilter` no longer count towards `maxCount` ([#1426](https://github.com/expressjs/multer/pull/1426))
 - Accept requests with exactly `limits.parts` parts; `LIMIT_PART_COUNT` now fires only when the limit is exceeded. If you set `parts` one higher to work around this, you can drop the extra one ([#1446](https://github.com/expressjs/multer/pull/1446))
 
 ## 2.3.0
--- a/README.md
+++ b/README.md
@@ -160,6 +160,9 @@ Accept an array of files, all with the name `fieldname`. Optionally error out if
 more than `maxCount` files are uploaded. The array of files will be stored in
 `req.files`.
 
+Files skipped by `fileFilter` do not count towards `maxCount`. Use
+`limits.files` to bound how many files are read from a request.
+
 #### `.fields(fields)`
 
 Accept a mix of files, specified by `fields`. An object with arrays of files
--- a/index.js
+++ b/index.js
@@ -46,8 +46,13 @@ Multer.prototype._makeMiddleware = function (fields, fileStrategy) {
         return cb(new MulterError('LIMIT_UNEXPECTED_FILE', file.fieldname, file.originalname))
       }
 
-      filesLeft[file.fieldname] -= 1
-      fileFilter(req, file, cb)
+      // Only count the file against the field's maxCount once the user's
+      // fileFilter has accepted it. A file skipped via cb(null, false) is
+      // never stored, so it must not consume a slot (see #1419).
+      fileFilter(req, file, function (err, includeFile) {
+        if (!err && includeFile) filesLeft[file.fieldname] -= 1
+        cb(err, includeFile)
+      })
     }
 
     return {
--- a/test/file-filter.js
+++ b/test/file-filter.js
@@ -18,6 +18,10 @@ function reportFakeError (req, file, cb) {
   cb(new Error('Fake error'))
 }
 
+function skipByOriginalName (req, file, cb) {
+  cb(null, file.originalname !== 'tiny0.dat')
+}
+
 describe('File Filter', function () {
   it('should skip some files', function (done) {
     var form = new FormData()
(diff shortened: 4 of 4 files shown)


MohammedAlkindi:
The counting change is right, but it removes the only default bound on how many files multer reads for a field.

With `upload.array('docs', 1)` and a filter that rejects everything, `main` stops at the first file with `LIMIT_UNEXPECTED_FILE`; this branch reads all 50 and finishes without error. `filesLeft` never decrements for rejected files, so nothing caps the count except `limits.files`, which is unset by default.

<details>
<summary>Node 24, 50 files on one field, fileFilter rejects all</summary>

```
main    err=LIMIT_UNEXPECTED_FILE  fileFilterCalls=1   stored=0
#1426   err=null                   fileFilterCalls=50  stored=0
```
</details>

Either cap reads per field, or say in the docs that `limits.files` is now the bound. An async fileFilter showed no over-limit race, since busboy serialises the file events. The suite is green on Windows apart from #1451.

UlisesGascon:
Went with the docs option: `maxCount` counts accepted files, and `limits.files` is the bound on how many files are read (README updated in this PR). Thanks for measuring it @MohammedAlkindi!

UlisesGascon (approved):

Issue #1205, opened 2023-04-26
Second request hangs after upload fail into filter function


danilomourelle:
I'm using multer to receive files just to send it to another service, so I'm using memoryStorage. I'm also filtering files by mime type and trying to use a custom error handling because I'm using a middleware error handling as last middleware. So I have the following in code

```typescript
// filter.ts
const fileFilter = (_request: Request, file: Express.Multer.File, callback: FileFilterCallback) => {
  const { mimetype } = file;
  const isPDF = mimetype === "application/pdf";
  const isJPEG = mimetype === "image/jpeg";
  const isPNG = mimetype === "image/png";

  if (isPDF || isJPEG || isPNG) {
    // eslint-disable-next-line unicorn/no-null
    callback(null, true);
  } else {
    logger.warn("Multer Filter - File sent with wrong extension", { mimetype });
    callback(new PreconditionFailed("Invalid file type, only documents PDF or images JPEG/PNG is allowed with a maximum size of 5 MB"));
  }
};

export default fileFilter;
```

and

```typescript
import fileFilter from "@services/multer/filter";
import buildMulterErrorMessage from "./buildMulterErrorMessage";

const FIVE_MB_IN_BYTES = 5_000_000;

const storage = multer.memoryStorage();

const upload = multer({
  storage,
  fileFilter,
  limits: { fileSize: FIVE_MB_IN_BYTES },
}).fields([
  { name: "FILE_01", maxCount: 1 },
  { name: "FILE_02", maxCount: 1 },
]);

const factory = (request: Request, response: Response, next: NextFunction) => {
   upload(request, response, (error) => {
    const isMulterError = error instanceof multer.MulterError;

    if (isMulterError) {
      const errorMessage = buildMulterErrorMessage(error);
      return next(new BadRequest(errorMessage));
    }
 
    return next(error);
  });
};

export default factory ;
```

I call this factory on routes middleware like `app.post('/file', factory, module)` and it works fine, but if we have a error on filter, the PreconditionError is sent as response, and if I retry the request, this second is hanging. It could be any other endpoint on the API. rigth now I have to manually cancel this second and do a third, which works, and so on, it's keeping this on/off behavior.

### Edit 1

I found out that the problem is to throw an error in fileFilter callback function. If I replace that by `callback(null, false)` and throw the error in `next()´ into `upload` function, it does not hang. But in this way I cant figure out if any file was filtered.

### Edit 2

[...]

Issue #1093, opened 2022-05-20
Field upload throw error when filter reject then file.


AlexAdvent:
I am uploading two image using `.fields([{ name: 'certificateImage', maxCount: 1 }, { name: 'addressProofImage', maxCount: 1 }]); ` but it is throwing error when filefilter reject the file

error
```
  if (this.req.files[placeholder.fieldname].length === 1) {
                                                ^

TypeError: Cannot read properties of undefined (reading 'length')
```

Multer throw error at line 49 in file-appender.js. req.files[placeholder.fieldname] is showing undefined.
```
case 'OBJECT':
      if (this.req.files[placeholder.fieldname].length === 1) {
        delete this.req.files[placeholder.fieldname]
      } else {
        arrayRemove(this.req.files[placeholder.fieldname], placeholder)
      }
      break

```

My code

```
const multer  = require('multer');
const path = require('path');
const Master = require('../models/setting/master'); 

var storage = multer.diskStorage({
    destination: function (req, file, cb) {
      console.log('type',req.body.type);
      cb(null, "uploads/images/"+file.fieldname+"/");
    },
    filename: function (req, file, cb) {
      cb(
        null,
        file.fieldname + "-" + Date.now() + path.extname(file.originalname)
      );
    },
  });

const maxSize = 1 * 1024 * 1024; // for 1MB

var upload = multer({
    storage: storage,
    fileFilter: async (req, file, cb) => {

      var ext = path.extname(file.originalname);
      console.log("ext", ext);

      // validate extension from store mongodb ext
      let data = await Master.findOne({});
      
      let validImageExtensions = data.validImageExtensions;
      console.log("validImageExtensions", validImageExtensions);

      if (validImageExtensions.indexOf(ext.substring(1)) === -1) {
        cb(null, false); 
        console.log("invalid extension");
        return cb(new Error("Only " + validImageExtensions + " are allowed with maxsize 1MB"));  // **occouring issue here**
      }else{
        console.log("valid extension");
        cb(null, true);
      }
    },
    limits: { fileSize: maxSize },
  });

module.exports = upload;
```

when extension doesn't and reject the file multer throw error.

LinusU:
Very interesting! Would you be able to submit a PR with a failing test case?

geminigeek:
hi,
Having same issue while the uploaded file size is larger than the fileSize value

Lone-Wolf17:
Also experiencing this issue.

andreicimpoesu:
[...]

PR #1427, opened 2026-07-02
fix: prevent process crash when fileFilter calls its callback more than once

### LINKED ISSUE #950: Length is undefined in "if (this.req.files[placeholder.fieldname].length === 1)" when uploading file with invalid extension
When I try to upload file with the extension not specified in my fileFilter object I get this error:
```
if (this.req.files[placeholder.fieldname].length === 1) {
TypeError: Cannot read property 'length' of undefined
```
Worth to note is that everything works fine if the uploaded files match the condition in fileFilter. Also the same code works if I use single() and upload only one file instead of using fields().

**Code:**
```
const upload = multer({
  storage,
  fileFilter: (req, file, cb) => {
    if (file.mimetype.includes("image")) {
      cb(null, true);
    } else {
      cb(null, false);
      return cb({ code: "INVALID_EXTENSION" });
    }
  },
}).fields([
  { name: "logo", maxCount: 1 },
  { name: "banner", maxCount: 1 },
  { name: "image", maxCount: 5 },
]);
```

spokodev:
Fixes #950.

With `upload.fields()`, if a `fileFilter` invokes its callback more than once and does so asynchronously, Multer crashes the whole Node process with an uncaught exception:

```
TypeError: Cannot read properties of undefined (reading 'length')
    at FileAppender.removePlaceholder (lib/file-appender.js)
```

A placeholder is inserted synchronously when the file event fires, then removed asynchronously from the `fileFilter` callback. When the callback runs twice, `removePlaceholder` runs twice for the same field: the first call sees length 1 and deletes `req.files[fieldname]`; the second call reads `.length` of the now-missing entry and throws on a later tick, where there is no surrounding try/catch, so the process dies. This has been reported repeatedly (#950, #1093, #1205).

Only the `OBJECT` strategy (`upload.fields()`) is affected; the `ARRAY` strategy (`upload.array()`) already tolerates a double callback because `arrayRemove` no-ops on a missing entry.

## Fix

Break out of the `OBJECT` branch when the field entry is already gone, matching the null-safety the `ARRAY` branch already has. The single-callback path and behavior are unchanged.

## Testing

Added a test to `test/file-filter.js`: a `.fields()` parser whose `fileFilter` calls the callback twice via `setImmediate` (the asynchronous case). It crashes the process on the current code and passes with the fix. Full suite stays green (81 passing).

### DIFF
--- a/lib/file-appender.js
+++ b/lib/file-appender.js
@@ -42,14 +42,13 @@ FileAppender.prototype.removePlaceholder = function (placeholder) {
     case 'NONE': break
     case 'VALUE': break
     case 'ARRAY': arrayRemove(this.req.files, placeholder); break
-    case 'OBJECT':
-      if (!this.req.files[placeholder.fieldname]) break
-      if (this.req.files[placeholder.fieldname].length === 1) {
-        delete this.req.files[placeholder.fieldname]
-      } else {
-        arrayRemove(this.req.files[placeholder.fieldname], placeholder)
-      }
+    case 'OBJECT': {
+      var files = this.req.files[placeholder.fieldname]
+      if (!files) break
+      arrayRemove(files, placeholder)
+      if (files.length === 0) delete this.req.files[placeholder.fieldname]
       break
+    }
   }
 }
 
--- a/lib/file-appender.js
+++ b/lib/file-appender.js
@@ -43,6 +43,7 @@ FileAppender.prototype.removePlaceholder = function (placeholder) {
     case 'VALUE': break
     case 'ARRAY': arrayRemove(this.req.files, placeholder); break
     case 'OBJECT':
+      if (!this.req.files[placeholder.fieldname]) break
       if (this.req.files[placeholder.fieldname].length === 1) {
         delete this.req.files[placeholder.fieldname]
       } else {
--- a/test/file-filter.js
+++ b/test/file-filter.js
@@ -53,4 +53,27 @@ describe('File Filter', function () {
       done()
     })
   })
+
+  it('should not crash when fileFilter invokes the callback more than once', function (done) {
+    function rejectThenError (req, file, cb) {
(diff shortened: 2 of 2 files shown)


MohammedAlkindi:
The guard stops the crash, but the same double callback still silently drops an accepted file.

`removePlaceholder` is not idempotent under the OBJECT strategy. With `.fields([{ name: 'docs', maxCount: 2 }])`, two files on `docs`, and a filter that calls `cb(null, false)` twice for the second one, it runs twice for a single placeholder. The first call takes the `arrayRemove` branch and leaves one real file behind. The second sees `length === 1` and runs `delete this.req.files[placeholder.fieldname]` (`lib/file-appender.js:47-48`), dropping that stored file. Nothing is raised, so the caller just sees the field gone.

The `length === 1` branch deletes the whole field without checking that the placeholder being removed is the element still in it. Deleting only once the array is actually empty covers both cases:

```js
case 'OBJECT': {
  var files = this.req.files[placeholder.fieldname]
  if (!files) break
  arrayRemove(files, placeholder)
  if (files.length === 0) delete this.req.files[placeholder.fieldname]
  break
}
```

Your new test still passes with that change, and the suite's failing set is unchanged either way: 67 passing / 13 failing on both sides, the 13 being pre-existing Windows path failures on my box rather than anything this PR touches.

spokodev:
Great catch, thanks @MohammedAlkindi. You're right that the crash guard left the data-loss in place. I applied your idempotent version of the OBJECT branch (arrayRemove first, delete the field only when the array is empty) and credited you as co-author on the commit. I also added a regression test: `.fields([{ name: 'docs', maxCount: 2 }])` with a duplicate `cb(null, false)` on the second file, asserting the first accepted file survives. It failed on the previous commit and passes now; full suite is green (82 passing) and `standard` is clean.

PR #1428, opened 2026-07-10
docs: add Express Router upload example

### LINKED ISSUE #596: Proposed addendum to Readme.md for Express Router users?
I really struggled to use this library with Express and the Router function.  Can I propose adding the following example, at the bottom of the usage block on the Readme.md file?

> For anybody using Express and Router()...

```
var router = express.Router();

var storage = multer.diskStorage({
    destination: (req, file, cb) => {
      cb(null, directory)
    },
    filename: (req, file, cb) => {
      cb(null, file.originalname)
    }
});
var uploadMulter = multer({storage: storage});

router.post('/uploadFiles', uploadMulter.array("input_element_name"),  function(req,res, next) {
	console.log(req.body); //form fields
	console.log(req.file); //form files
    res.status(204).end();
});
```
Pull request ?  Can I , can I ??

vibhor-aggr:
## Summary

Adds a concise README example showing Multer middleware used with an `express.Router()` route.

This addresses the confusion raised in #596 by showing that the same `upload.array(...)` middleware pattern used on `app.post(...)` can be mounted directly on a router route.

Fixes #596.

## Verification

- `npm test` (80 passing)
- `npm run lint`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,24 @@ app.post('/cool-profile', uploadMiddleware, function (req, res, next) {
 })
 ```
 
+The same upload middleware can be used with an `express.Router()` instance:
+
+```javascript
+const express = require('express')
+const multer  = require('multer')
+
+const router = express.Router()
+const upload = multer({ dest: 'uploads/' })
+
+router.post('/uploadFiles', upload.array('documents'), function (req, res, next) {
+  // req.files is array of `documents` files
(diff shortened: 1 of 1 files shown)


UlisesGascon:
Thanks @vibhor-aggr but router usage is the same as `app.post(...)`, nothing multer-specific to document, so closing it to keep a small readme :+1:

PR #1435, opened 2026-07-22
docs: updated README's note with a note blockquote

fdipilla:
Updated README's note with a note blockquote. This change makes the note more readable and harder to miss (don't ask me how I know).

The README goes from this:

image

to this:

image


UlisesGascon:
Thanks @fdipilla for helping the project! Unfortunately we can't take this one: the GitHub alert syntax (`> [!NOTE]`) isn't supported by npm's README renderer, and this file is what's shown on the npm package page — so npm users would see the literal `[!NOTE]` text inside a blockquote. We'll keep the plain `**NOTE**:` for now.

PR #1446, opened 2026-08-02
fix: allow exactly limits.parts parts

### LINKED ISSUE #1418: Parts Limit is misleading
### Description

The parts limit is advertised as files + fields, but multer faithfully raises an error as soon as parts limit is hit. For the other limits, they reason after you exceed the limit. This seems to have to do with busboy emitting a parts limit event as soon as its hit, as well it counting the final boundary as a part. So if you're upgrading the latest multer and are thinking "I should set partsLimit as files + fields for security reasons", you'll realize with testing that you need to add 1 to account for the boundary that busboy counts.

### Expectations

I would have expected an error when parts limit has been exceeded, which to me means files + fields

deepakganesh78:
busboy emits `partsLimit` as soon as the part count reaches `limits.parts`, so a request with exactly the allowed number of parts was rejected with `LIMIT_PART_COUNT`. This passes `parts + 1` to busboy, the same way `fileSize` is handled since #1407, so the error fires only when the limit is exceeded; busboy still stops parsing at that point. The README workaround from #1430 is removed since it no longer applies.

Rebased onto `main` as a single commit; tests cover exactly-at-limit (accepted) and one-over (rejected).

Closes #1418

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+- Accept requests with exactly `limits.parts` parts; `LIMIT_PART_COUNT` now fires only when the limit is exceeded. If you set `parts` one higher to work around this, you can drop the extra one ([#1446](https://github.com/expressjs/multer/pull/1446))
+
 ## 2.3.0
 
 - Fix [CVE-2026-77078](https://www.cve.org/CVERecord?id=CVE-2026-77078) ([GHSA-wc9g-mqfw-jrwm](https://github.com/expressjs/multer/security/advisories/GHSA-wc9g-mqfw-jrwm))
--- a/README.md
+++ b/README.md
@@ -278,10 +278,6 @@ Key | Description | Default
 `fieldNestingDepth` | Max number of nesting levels for field names (e.g. `a[b][c]` has 2 levels) | Infinity
 `fieldArrayIndexLimit` | Max numeric array index accepted inside a field name (e.g. `a[3]` uses index 3) | Infinity
 
-The `parts` limit is triggered when busboy reaches the configured number of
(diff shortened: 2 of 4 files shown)


MohammedAlkindi (commented):
Ran this on Windows 11, Node 24.18.0, against `main` at `4203ccd`.

**The off-by-one is real and the fix discriminates.** This suite is nondeterministic on Windows (base runs vary 70–71 passing / 12–13 failing on the same commit), so raw counts are useless — I compared failing-test *sets* with the lib reverted against the PR head, same test files both sides:

- Fails **only** with the old lib: `Error Handling should allow parts up to the limit` — the PR's new test, i.e. a request with exactly `parts: 2` parts wrongly aborts with `LIMIT_PART_COUNT` today, exactly as #1418 describes.
- Fails **only** on the PR: **nothing**. (The other difference in the reverted run was the known-flaky `Select Field` before-all hook, which comes and goes on identical commits here.)

Run directly, the three parts-limit tests pass on the PR head.

**On the implementation:** both the `field` and `file` handlers route through `incrementPartCount()`, so the count covers all part kinds; the `typeof/isFinite` guard means no behaviour change when `parts` is unset; and the object spread matches existing usage in this file (lines ~259/269), so no style drift.

**One semantic shift worth a sentence in the description:** since `parts` is now stripped from the limits busboy sees, busboy will begin parsing the N+1th part — headers included — before multer aborts, where previously busboy stopped emitting at its own boundary. For well-formed requests that's invisible; for a hostile stream it means the abort lands one part-header later than before. I don't think it matters given `LIMIT_PART_COUNT` still fires and the request is torn down, but it's the kind of thing a future reader will wonder about, and the PR description is the right place to say it was considered.

Also worth noting for reviewers: `busboyLimits()` returns the original object untouched when `parts` isn't set — no copy — so there's no per-request allocation cost in the common path. Nice touch.

UlisesGascon (approved):

PR #1447, opened 2026-08-04
Update busboy link pointing to nonexistent section

krzysdz:
Busboy readme file structure changed when 1.0.0 was released (in https://github.com/mscdex/busboy/commit/54a86838c15bba1fc78eebdfa3c6a986a5e57dd9 to be exact).

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ memory storage is used.
 
 ### `limits`
 
-An object specifying the size limits of the following optional properties. Multer passes this object into busboy directly, and the details of the properties can be found on [busboy's page](https://github.com/mscdex/busboy#busboy-methods).
+An object specifying the size limits of the following optional properties. Multer passes this object into busboy directly, and the details of the properties can be found on [busboy's page](https://github.com/mscdex/busboy#exports).
 
 The following integer values are available:
 


krzysdz:
By the way, it might be worth noting in the readme that `fieldNestingDepth` is a limit enforced by multer and not busboy:

https://github.com/expressjs/multer/blob/2e2af08157c66cbbc76539ccbd8869097a0c8feb/README.md?plain=1#L271

It took me a moment to realise that the details of the `fieldNestingDepth` property ca**n't** be found on busboy's page.

krzysdz:
Speaking of `fieldNestingDepth`, this seems to be incorrect:

https://github.com/expressjs/multer/blob/2e2af08157c66cbbc76539ccbd8869097a0c8feb/README.md?plain=1#L318

`a[b][c]` should work with `fieldNestingDepth: 2` as suggested [earlier in the readme](https://github.com/expressjs/multer/blob/2e2af08157c66cbbc76539ccbd8869097a0c8feb/README.md?plain=1#L284).



Additionally, shouldn't it be set to the **maximum** depth that field names require?

UlisesGascon (approved):

UlisesGascon:
I agree! We need to work a bit on `fieldNestingDepth` :+1:

PR #1448, opened 2026-08-07
fix: fallback to error code when no message mapping exists

### LINKED ISSUE #659: MulterError class doesn't allow custom messages.
https://github.com/expressjs/multer/blob/a04f29376aba2ecf7062ff6a00749a4afb39db78/lib/multer-error.js#L16

I didn't expect this behavior. I expected it to create a custom message that I provided.
`File type not supported!`
I am checking meme types using the filter. I thought I could use the `MulterError` class to nicely catch and organize its errors.

sahilraut191685:
Closes #659

Currently `errorMessages[code]` returns `undefined` if the code isn't recognized, so `this.message` silently becomes `undefined` with no useful information.

This adds a fallback so the code itself is used as the message when there's no mapping:

this.message = errorMessages[code] || code

Small, safe first step — happy to follow up with tests or explore a fuller custom-message fix if maintainers think that's worth pursuing.

### DIFF
--- a/lib/multer-error.js
+++ b/lib/multer-error.js
@@ -15,7 +15,7 @@ var errorMessages = {
 function MulterError (code, field) {
   Error.captureStackTrace(this, this.constructor)
   this.name = this.constructor.name
-  this.message = errorMessages[code]
+  this.message = errorMessages[code] || code
   this.code = code
   if (field) this.field = field
 }
--- a/lib/multer-error.js
+++ b/lib/multer-error.js
@@ -15,7 +15,7 @@ var errorMessages = {
 function MulterError (code, field) {
   Error.captureStackTrace(this, this.constructor)
   this.name = this.constructor.name
-  this.message = errorMessages[code] || code
+  this.message = errorMessages[code] || `Unknown error: ${code}`
   this.code = code
   if (field) this.field = field
 }
--- a/test/error-handling.js
+++ b/test/error-handling.js
@@ -11,19 +11,16 @@ var FormData = require('form-data')
 var http = require('http')
 var net = require('net')
 
-function withLimits(limits, fields) {
+function withLimits (limits, fields) {
   var storage = multer.memoryStorage()
   return multer({ storage: storage, limits: limits }).fields(fields)
 }
 
 describe('Error Handling', function () {
-
-
   it('should use a fallback message when the code is not mapped', function () {
     var err = new MulterError('SOME_NEW_CODE')
     assert.strictEqual(err.message, 'Unknown error: SOME_NEW_CODE')
   })
-
 
   it('should be an instance of both `Error` and `MulterError` classes in case of the Multer\'s error', function (done) {
     var form = new FormData()
@@ -202,7 +199,7 @@ describe('Error Handling', function () {
   it('should report errors from storage engines', function (done) {
     var storage = multer.memoryStorage()
 
-    storage._removeFile = function _removeFile(req, file, cb) {
+    storage._removeFile = function _removeFile (req, file, cb) {
       var err = new Error('Test error')
       err.code = 'TEST'
       cb(err)
@@ -332,7 +329,7 @@ describe('Error Handling', function () {
         })
       }, 8000)
 
-      function finish(err) {
+      function finish (err) {
         if (finished) return
         finished = true
         clearTimeout(timeout)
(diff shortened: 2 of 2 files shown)


MohammedAlkindi (commented):
Ran this on Windows 11 (10.0.26200), Node 24.18.0, against `upstream/main` at `4203ccd`.

**The change is right.** `errorMessages[code]` yields `undefined` for any code without a mapping, so `new MulterError('SOME_NEW_CODE').message` is `undefined` and the error prints as `MulterError: undefined` — which is worse than useless in a log. Falling back to `code` at least names what happened.

**Test evidence — and a caveat that matters more than the numbers.** The suite is **not deterministic on Windows**, so I ran each side twice:

| | run 1 | run 2 |
|---|---|---|
| `main` @ `4203ccd` | 66 passing / 13 failing | **65 passing / 15 failing** |
| this PR | 66 passing / 13 failing | 66 passing / 13 failing |

The base varies by two failures **between runs of the same commit**, so the PR sits inside the noise band and **no regression is attributable to it**. I'm reporting the variance rather than a single clean delta because a single run here would have supported whatever conclusion I ran into first — my first two passes happened to show 71/12 then 70/13 on the *same* commit, which is exactly the trap.

(The failures themselves look like the pre-existing Windows issues — the `Unicode` hooks and the `dest`/rename tests. Not this PR's business.)

**Two things I'd raise:**

1. **No test.** This is a one-line behaviour change to a public class — `MulterError` is exported and users match on it. A three-line assertion (`new MulterError('NOT_A_REAL_CODE').message === 'NOT_A_REAL_CODE'`) would pin the fallback so it can't silently regress. Given the suite's flakiness on some platforms, a deterministic unit test here is cheap insurance.

2. **It is technically a public-API behaviour change.** Anything doing `if (err.message === undefined)` to detect an unmapped code would now see a string. I think that's fine — nobody sensible depends on `undefined` — but it's worth a line in the release notes rather than landing silently as a pure bugfix.

One design thought, take it or leave it: `errorMessages[code] || code` makes `message` and `code` identical for unmapped codes, so a log line reads `MulterError: LIMIT_WEIRD_THING` with no indication the code was unrecognised. Something like `` errorMessages[code] || `Unknown error: ${code}` `` would tell a maintainer reading a bug report that the code fell through the mapping. Slightly noisier, considerably more diagnostic.

---

*Disclosure: I used an AI assistant while reviewing. The four suite runs, the flakiness measurement, and the base/PR comparison above were executed by me on the Windows machine described.*

sahilraut191685:
Thank you for the detailed review and for validating the change across multiple runs. I agree that adding a deterministic unit test for the unmapped error-code fallback is appropriate. I’ll add the test and push the update to the PR.

I also agree that the public API behavior change should be documented in the release notes.

sahilraut191685:
Added the requested regression test for the unmapped error-code fallback. The test passes successfully, and I’ve pushed the changes to the PR.

MohammedAlkindi:
The fallback and its test are right: `npx mocha --exit --check-leaks test/error-handling.js` on `0d10d3e` gives 19 passing, including `should use a fallback message when the code is not mapped`.

Lint regressed though. `npm run lint` exits 1 here and 0 on `origin/main`:

```
test/error-handling.js:14:20   Missing space before function parentheses
test/error-handling.js:19:40   Block must not be padded by blank lines
test/error-handling.js:21:1    More than 1 blank line not allowed
test/error-handling.js:27:1    More than 1 blank line not allowed
test/error-handling.js:205:47  Missing space before function parentheses
test/error-handling.js:335:22  Missing space before function parentheses
test/error-handling.js:344:26  Missing space before function parentheses
test/error-handling.js:459:24  Missing space before function parentheses
```

Five are on functions this change did not need to touch. `npm run lint:fix` clears all eight.

sahilraut191685:
Fixed the fallback error message for unknown error codes.
Also fixed the lint issues in the test file.
Tests and lint are passing now.

MohammedAlkindi:
Confirmed clean at `f6720ce`. `npm run lint` exits 0, and `npx mocha --exit --check-leaks test/error-handling.js` gives 19 passing.

Net diff against `main` is now `lib/multer-error.js` +2/-1 and `test/error-handling.js` +9/-2, so the reformatting of the untouched functions is gone and only the fallback plus its regression test remain.

sahilraut191685:
Thank you for the thorough review and for confirming the fix — appreciate you taking the time to verify everything.

UlisesGascon (approved):
Thanks @sahilraut191685!

PR #1452, opened 2026-08-16
fix(test): use path.sep for cross-platform path assertions

### LINKED ISSUE #1451: Failing test on Windows: functionality.js hardcodes forward slash
## Environment information

**Version**: 2.2.0
<!-- 

```json
"dependencies": {
  "express": "4.18.2"
}
-->

**Platform**: WINDOWS
<!-- 
UNIX: output of uname -a
Windows: output of "$([Environment]::OSVersion.VersionString) $(('x86', 'x64')[[Environment]::Is64BitOperatingSystem])" in PowerShell console 
-->

**Node.js version**: 22.0.0
<!-- 
Output of node -v.
-->

when running tests on Windows, should rename the destination directory to a different directory fails because assert.ok(req.files[0].path.indexOf('/testforme-') >= 0) hardcodes a forward slash, which fails against Windows backslash paths.

### PR DESCRIPTION
**Pull request by kilisamemarisaaa**
## Summary

Fixes #1451.

The `should rename the destination directory to a different directory` test in `test/functionality.js` hardcodes `/` as the path separator in its assertions:

```js
assert.ok(req.files[0].path.indexOf('/testforme-') >= 0)
assert.ok(req.files[1].path.indexOf('/testforme-') >= 0)
```

On Windows, `path` uses `\` as the separator, so `indexOf('/testforme-')` returns `-1` and the test fails even though the upload directory was correctly created by `fs-temp`.

## Fix

Replace the hardcoded `'/'` with `path.sep` so the assertion holds on both POSIX and Windows:

```js
assert.ok(req.files[0].path.indexOf(path.sep + 'testforme-') >= 0)
assert.ok(req.files[1].path.indexOf(path.sep + 'testforme-') >= 0)
```

A `require('path')` is added at the top of the file. No production code is changed — only the test assertion.

## Test plan

- [x] Ran `npx mocha --exit --timeout 30000 test/functionality.js` on Windows:
  - `should rename the destination directory to a different directory` ✔ passes (was failing before on Windows)
  - `should rename the uploaded file` ✔
  - `should ensure all req.files values (single-file per field) point to an array` ✔
  - `should ensure all req.files values (multi-files per field) point to an array` ✔
- [x] Diff is minimal: 1 file changed, 4 insertions(+), 2 deletions(-)
- [x] No production code (`lib/`, `storage/`, `index.js`) touched

## Notes

- A separate, pre-existing Windows environment issue exists in this repo: `test/files/small0.dat` is a text fixture with LF line endings, but the repo has no `.gitattributes` pinning it to binary/LF. Git's autocrlf on Windows converts LF to CRLF, inflating the file from 1778 to 1803 bytes, which causes the unrelated `should upload the file to the dest dir` test to fail on `assert.strictEqual(util.fileSize(req.file.path), 1778)`. That is **out of scope** for #1451 (which is specifically about the path separator) and is not addressed by this PR to keep the diff focused.

### DIFF
--- a/test/functionality.js
+++ b/test/functionality.js
@@ -2,6 +2,8 @@
 
 var assert = require('assert')
 
+var path = require('path')
+
 var util = require('./_util')
 var multer = require('../')
 var temp = require('fs-temp')
@@ -129,8 +131,8 @@ describe('Functionality', function () {
     util.submitForm(parser, form, function (err, req) {
       assert.ifError(err)
       assert.strictEqual(req.files.length, 2)
-      assert.ok(req.files[0].path.indexOf('/testforme-') >= 0)
-      assert.ok(req.files[1].path.indexOf('/testforme-') >= 0)
+      assert.ok(req.files[0].path.indexOf(path.sep + 'testforme-') >= 0)
+      assert.ok(req.files[1].path.indexOf(path.sep + 'testforme-') >= 0)
       done()
     })
   })


MohammedAlkindi (commented):
Ran this on a real Windows 11 box (Node 24.18), since the fork-PR test workflow here is approval-gated and hasn't run:

- upstream `main` (4203ccd): **71 passing, 12 failing**
- this branch (c48616e): **72 passing, 11 failing**

The only change in the failing *set* is `Functionality > should rename the destination directory to a different directory` — exactly the test #1451 reports — which fails on `main` on the hardcoded `'/testforme-'` assertion and passes here. No other test changes state in either direction.

The fix itself is right for what the assertion means: disk storage produces `...\testforme-<random>` on Windows, so `path.sep + 'testforme-'` checks the same "directory boundary + prefix" property on both platforms, and `indexOf` with a literal separator has no escaping pitfalls.

For context on the remaining 11: they are a different Windows breakage — Git rewriting the text fixtures to CRLF on checkout, so the byte-length assertions drift (`1803 !== 1778`, `128 !== 122`). That class is what #1450 addresses; it's untouched by (and doesn't interfere with) this PR.

UlisesGascon (approved):

PR #1458, opened 2026-08-27
fix: add flush option to disk storage to fsync files before completion

### LINKED ISSUE #1381: uploaded files are not flushed to disk

kilisamemarisaaa:
Fixes #1381

## Root cause

`storage/disk.js` opens the output stream with `fs.createWriteStream(finalPath)` and calls `cb(null, …)` directly on the stream's `'finish'` event — there is no `fsync`/`fdatasync` barrier anywhere in the write path. `'finish'` only means the data reached the OS page cache; if the process or the machine crashes before the cache is written back, the uploaded file can be lost or left corrupt even though the handler already observed the upload completing.

**Note on the approach suggested in the issue:** the `flush: true` option mentioned in #1381 only exists on `filehandle.createWriteStream()` (Node 20.10+); the `fs.createWriteStream()` used here has no such option. This PR uses the mechanism that works on every supported Node version (>= 10.16.0) and platform: after the stream finishes, call `fs.fsync()` on the stream's file descriptor.

## Fix

- `storage/disk.js` — new opt-in `flush` option on `multer.diskStorage()`. When `flush: true`, the `'finish'` handler waits for `fs.fsync(outStream.fd, done)` before resolving the file, so the data is durable before the handler runs. Default stays disabled → no behavior change and no per-upload fsync cost for existing users.
- `test/disk-storage.js` — two new tests: fsync is **not** called by default, and is called **exactly once** when `flush: true` (the file itself is still verified on disk via the existing read-back assertions). Zero new dependencies, no test-framework changes.
- `README.md` — document the `flush` option in the DiskStorage section.

## Verification

- `npm run lint` (standard): pass
- `npm test`: **102 passing, 0 failing** (includes the 2 new cases + the pre-existing 9 disk-storage tests), on Node v24.12.0
- Full CI matrix (ubuntu + windows × Node 10–26) will exercise the fsync path cross-platform.

## Why opt-in rather than always-on

Forcing an `fsync` on every upload adds latency and storage wear. Keeping it opt-in preserves existing behavior and performance while giving durability to apps that need it. If the maintainers would rather default it to on (or make it a top-level multer option instead), I'm happy to adjust.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- Add opt-in `flush` option to `DiskStorage` to fsync files before the callback runs ([#1458](https://github.com/expressjs/multer/pull/1458))
 - Accept a function for `limits`, called with the request, to set limits per request ([#1133](https://github.com/expressjs/multer/pull/1133))
 - Files skipped by `fileFilter` no longer count towards `maxCount` ([#1426](https://github.com/expressjs/multer/pull/1426))
 - Accept requests with exactly `limits.parts` parts; `LIMIT_PART_COUNT` now fires only when the limit is exceeded. If you set `parts` one higher to work around this, you can drop the extra one ([#1446](https://github.com/expressjs/multer/pull/1446))
--- a/README.md
+++ b/README.md
@@ -251,7 +251,9 @@ null as the first param), refer to
 disk with an `fsync` call before the upload handler runs, which prevents data
 loss if the process or the system crashes while the data is still sitting in
 the operating system's page cache. It is disabled by default because forcing a
-disk flush on every upload has a performance cost.
+disk flush on every upload has a performance cost. Note that this syncs the
+file contents only; if the new directory entry must survive a crash as well,
+fsync the destination directory yourself.
 
 #### `MemoryStorage`
 
--- a/README.md
+++ b/README.md
@@ -217,8 +217,9 @@ const storage = multer.diskStorage({
 const upload = multer({ storage: storage })
 ```
 
-There are two options available, `destination` and `filename`. They are both
-functions that determine where the file should be stored.
+There are three options available: `destination`, `filename` and `flush`. All of
+them except `flush` are functions that determine where the file should be
+stored.
 
 `destination` is used to determine within which folder the uploaded files should
 be stored. This can also be given as a `string` (e.g. `'/tmp/uploads'`). If no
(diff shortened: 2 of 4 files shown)


kilisamemarisaaa:
Pushed a follow-up commit fixing the CI failure on the older Node legs.

**What failed:** `should call fsync once when flush is enabled` failed on Node 10/11/12/15 (and 13/14/16–21 on Windows) with `fsyncCalls` being `0` instead of `1`. Root cause: `outStream.fd` is already `null` by the time `'finish'` fires on those versions, so the `that.flush && outStream.fd` guard silently skipped the flush — i.e. on exactly the versions where the guard mattered, the option did nothing.

**Fix:** the flush no longer depends on the write stream's descriptor lifetime. After `'finish'`, the file is reopened with `fs.open(finalPath, 'r+')`, `fs.fsync`'d, and closed; errors from either step are forwarded to the callback. `fsync` is a per-file operation, so this is equivalent to syncing the original descriptor and behaves the same on every supported Node version (>= 10.16.0) and platform.

Re-verified locally on Node v24.12.0: `npm run lint` clean, `npm test` **102 passing, 0 failing**. On CI both `flush option` tests now pass on every leg, including the Windows ones that were red before.

**On the remaining red Windows legs:** they fail on `orphan file cleanup on abort/malformed requests`, not on anything this PR touches, and they are pre-existing on `main`:

- my docs-only PR #1457 (zero source changes) fails on the same 12 Windows legs;
- PR #1456, which fixes exactly that orphan-cleanup bug, is green on all 40 checks including every Windows leg.

So `main` is currently red on `windows-latest` for Node < 22, this branch inherits it, and #1456 is the change that clears it. Those legs are marked `experimental: true` / `continue-on-error` in `ci.yml`. Every Ubuntu leg (Node 10–26), `lint`, `coverage`, `Analyze` and CodeQL are green here.

kilisamemarisaaa:
Rebased onto current main. The security fix in eef7444 (GHSA-qfvm-cv95-jqjf) rewrote `_handleFile` from `pipe` + `'finish'` to `pipeline`, so the flush step now runs in the pipeline completion callback instead of a `'finish'` handler. The by-path reopen approach is unchanged — it was already the right call, and it composes cleanly with the new `openStreams` bookkeeping since `fsync` happens after the pipeline (and thus the write stream) has finished. All tests and lint pass locally.

UlisesGascon (approved):
Awesome work @kilisamemarisaaa! I just pushed a small docs change 👍

PR #1459, opened 2026-08-28
docs: document StorageEngine abort contract (file.stream error/close semantics)

kilisamemarisaaa:
## Description

Follow-up to the maintainer direction in https://github.com/expressjs/multer/issues/1459#issuecomment-… : instead of adding a second teardown path (`file.outStream`), document the existing abort contract in `StorageEngine.md`.

## What changed

- `StorageEngine.md`: new paragraph documenting that on request abort / mid-upload failure, `file.stream` emits `error` followed by `close`; engines should prefer `stream.pipeline(file.stream, out, cb)` (the pattern `DiskStorage` uses since 2.3.0) — `pipeline` destroys the output stream (releasing handles such as open file descriptors) and calls the engine's `cb` with the error, so engines need neither `req.on('aborted')` listeners nor an `outStream` reference on the file object.
- The previously proposed code change (`file.outStream` handling in `lib/make-middleware.js`) and its test (`test/abort-outstream.js`) have been dropped from this PR.

## Test

Docs-only change; no runtime behavior modified.

### DIFF
--- a/StorageEngine.md
+++ b/StorageEngine.md
@@ -26,6 +26,13 @@ The file data will be given to you as a stream (`file.stream`). You should pipe
 this data somewhere, and when you are done, call `cb` with some information on the
 file.
 
+If your engine writes the file data to a stream (for example an
+`fs.createWriteStream`), you can expose that stream as `file.outStream`. When
+the request is aborted or fails mid-upload, multer will call `.destroy()` on it,
+releasing the underlying handle (such as an open file descriptor) without you
+having to listen for `req.on('aborted')` yourself. This is optional — multer
+only touches `file.outStream` when your engine sets it.
+
 The information you provide in the callback will be merged with multer's file object,
 and then presented to the user via `req.files`.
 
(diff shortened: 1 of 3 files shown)


UlisesGascon:
Thanks @kilisamemarisaaa! We're not adding a second teardown path for this. When the request is aborted, `file.stream` already emits `error` and `close`, so an engine that uses `stream.pipeline(file.stream, out, cb)` (as `DiskStorage` does since 2.3.0) gets its output stream destroyed and its callback called with the error, without multer knowing about the stream. That's the contract we want to document in StorageEngine.md rather than a new `file.outStream` property :thinking:

kilisamemarisaaa:
Thanks @UlisesGascon — direction accepted. Reworked this PR as docs-only per your contract framing: the `file.outStream` code change (and its test) are dropped; `StorageEngine.md` now documents that on abort/mid-upload failure `file.stream` emits `error` then `close`, and that engines using `stream.pipeline(file.stream, out, cb)` (as `DiskStorage` does since 2.3.0) get the output stream destroyed and `cb` called with the error — no `req.on('aborted')` listener or `outStream` reference needed. Forced push: b07eaaf.

Issue #1293, opened 2025-01-12
Test _removeFile function in multer's custom storage engine


AkshayKrJha:
I went through the docs here: [https://github.com/expressjs/multer/blob/master/StorageEngine.md](url)    
Here the logic of _handleFile is clear, but there's doubt about _removeFile function, like    

- How will it get new parameters from file?
- If we save uploaded file name during _handleFile function, would it be possible to retrieve that in _removeFile?    
- How to make _removeFile invoked and test all these using a console log there ?

official-burak:
Yes. `_handleFile`'s `cb(null, info)` is merged onto the same `file` object that `_removeFile(req, file, cb)` receives, so `file.path` / `file.filename` (whatever you returned) are already there.

Multer only calls `_removeFile` when the parse aborts after a file was stored (size limit, later part error, `fileFilter` calling `cb(err)`). A successful request does not invoke it. `cb(null, false)` just skips that file and will not call `_removeFile`.

To see it fire, store one file then fail the next:

```js
const upload = multer({
  storage: myStorage,
  fileFilter (req, file, cb) {
    if (file.originalname === 'keep.txt') return cb(null, true)
    cb(new Error('reject'))
  }
}).array('docs', 2)
```

POST `keep.txt` then `skip.txt` on `docs`. The first file is stored, the second errors, `_removeFile` runs with the `file` from `_handleFile`. A `console.log` in `_removeFile` is enough to confirm.

UlisesGascon:
Closed by https://github.com/expressjs/multer/pull/1460

PR #1460, opened 2026-08-28
test: cover storage engine _removeFile invocation semantics

kilisamemarisaaa:
## Summary

Adds regression tests for the storage engine `_removeFile` contract raised in #1293: `_handleFile`'s `cb(null, info)` is merged onto the same `file` object that `_removeFile(req, file, cb)` receives (`fileInfo = { ...file, ...info }` in `lib/make-middleware.js`), `_removeFile` is only invoked when the request aborts after a file was stored, and a successful request never calls it.

## Test cases (`test/_removefile.js`)

- **successful request**: `_removeFile` is never called (semantics anchor)
- **later file exceeds size limit**: `_removeFile` receives the merged file object — `path` returned from `_handleFile` is visible, plus `fieldname` / `originalname`
- **fileFilter errors on a later file**: previously stored files are cleaned up via `_removeFile`
- **memory storage engine**: `_removeFile` is invoked on the error cleanup path

Uses a minimal custom storage engine (pattern from `test/abort-outstream.js`) with captured call records, and `test/_util.js`'s `submitForm`.

## Verification

- `npm test` → **106 passing** (102 existing + 4 new), no regressions
- `npx standard test/_removefile.js` → clean

Refs #1293


UlisesGascon (approved):
Thanks @kilisamemarisaaa! I just pushed some changes into additional commits (file rename + test comments translations) 👍

PR #1465, opened 2026-08-29
feat: expose busboy defCharset, highWaterMark and fileHwm options

### LINKED ISSUE #320: expose busboy's options (highWaterMark/fileHwm/defCharset/preservePath)

UlisesGascon:
Passes busboy's `defCharset`, `highWaterMark` and `fileHwm` through `multer(opts)`, next to the existing `defParamCharset`, using busboy's option names. Nothing changes when they are not set — busboy keeps its defaults (covered by a test). Note: busboy does not validate the charset name; an unknown `defCharset` yields `undefined` field values.

Supersedes #488 and #1102 (authors credited as co-authors).

Closes #488
Closes #1102
Closes #320

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,9 @@ Key | Description
 `limits` | Limits of the uploaded data
 `preservePath` | Keep the full client-supplied path in `file.originalname` instead of just the base name
 `defParamCharset` | Default character set to use for values of part header parameters (e.g. filename) that are not extended parameters (that contain an explicit charset). Default: `'latin1'`
+`defCharset` | Default character set to use for text field values that do not declare one. Default: `'utf8'`
+`highWaterMark` | `highWaterMark` of the multipart parser stream. Default: busboy's default
+`fileHwm` | `highWaterMark` of each file stream (`file.stream`). Default: busboy's default
 
 In an average web app, only `dest` might be required, and configured as shown in
 the following example.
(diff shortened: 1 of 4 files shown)


bjohansebas (approved):
LGTM

kilisamemarisaaa (commented):
Independently verified on Windows with Node v24.12.0 at 345bfad6: npm test passes (123 passing, 2 pending) and npm run lint is clean. The new busboy-options coverage exercises default/custom defCharset, .any() propagation, fileHwm, and parser highWaterMark; I also traced both _makeMiddleware and any() option paths and found no correctness or compatibility issues. My account cannot submit an APPROVE review without explicit repository access, so this is recorded as a factual verification comment.

PR #1307, opened 2025-05-11
[FEAT] Google Cloud Functions Support

emregunel:
# Add streamHandler option for Google Cloud Functions support

## Problem

Multer currently uses `req.pipe(busboy)` to process multipart form data. This works well in most environments, but causes issues in Google Cloud Functions (GCF) where the request body is pre-processed and made available as `req.rawBody`. When Multer tries to pipe the request in GCF, it fails because the request stream has already been consumed.

## Solution

This PR adds a new `streamHandler` option that allows customizing how request data is fed to busboy. Instead of hardcoding `req.pipe(busboy)`, the middleware now accepts a custom function that can handle both standard Express environments and pre-processed environments like GCF.

## Implementation Details

1. Added a new `streamHandler` configuration option to the Multer constructor
2. Created a default stream handler that maintains the existing `req.pipe(busboy)` behavior
3. Updated the middleware to use the custom handler when provided
4. Modified the cleanup logic to only unpipe when using the default pipe-based handler
5. Added documentation explaining the new option and its usage
6. Added comprehensive tests for both default and custom handlers

## Example Usage

For Google Cloud Functions or similar environments where the request body is pre-processed:

```javascript
const multer = require('multer')

// Create a custom stream handler that works with pre-processed bodies
function gcfStreamHandler(req, busboy) {
  if (req.rawBody) {
    // In Google Cloud Functions, use the pre-processed body
    busboy.end(req.rawBody)
  } else {
    // Fall back to standard behavior in other environments
    req.pipe(busboy)
  }
}

// Create multer instance with custom stream handler
const upload = multer({ 
  storage: multer.memoryStorage(),
  streamHandler: gcfStreamHandler
})

// Use the middleware as usual
app.post('/upload', upload.single('file'), (req, res) => {
  res.send('File uploaded successfully')
})
```

## Benefits

1. **Compatibility**: Makes Multer work seamlessly in Google Cloud Functions and similar environments
2. **Flexibility**: Provides a customization point for how request data is fed to busboy
3. **Backwards compatibility**: Default behavior remains unchanged
4. **Minimal changes**: Small, targeted modification to the codebase

## Testing

Added new tests that verify:

1. The default behavior continues to work as expected
2. The custom stream handler works properly with simulated GCF-style pre-processed bodies

All existing tests continue to pass.

## Documentation

Added the new `streamHandler` option to the README.md with example usage for Google Cloud Functions environments.

## Related Issues

[...]

PR #1466, opened 2026-08-29
feat: add streamHandler option to feed busboy from pre-consumed bodies

### LINKED ISSUE #572: empty req.body and req.file when run in firebase hosting

### LINKED ISSUE #1189: Add  configuration parameter to support Google Cloud Functions
Multer calls `req.pipe(busboy)` to stream in the request body for processing:
https://github.com/expressjs/multer/blob/07c9e84477e19f6f5016330b0c37c7cfbf085b8d/lib/make-middleware.js#L177

However, on Google Cloud Functions (GCF), it wraps Express to pre-stream the entire request body into a custom `res.rawBody` property. This causes Multer to break on that runtime. Here is some background discussion:
https://stackoverflow.com/a/47319614/1237919

It would be a very simple solution to instead call the following when running on GCF:
```js
busboy.end(req.rawBody)
```

This could be supported by either:
1) adding a configuration parameter specifically for GCF.
2) adding a configuration parameter function to override the reading in of data however the caller wants.

I would be happy to submit a pull request for this. Would you be willing to accept it? Do you prefer one of these 2 options?

Thanks.

UlisesGascon:
Adds an optional `streamHandler(req, busboy)` to `multer(opts)`. By default multer keeps doing `req.pipe(busboy)`; platforms that consume the request body before Express runs (Google Cloud Functions / Firebase Functions expose it as `req.rawBody`) can feed busboy themselves — the README shows the `rawBody` snippet. Nothing changes when the option is not set. Tests cover a request whose body is only available in `req.rawBody`, including the error path, and `.any()`.

Based on #1307, with the `rawBody` case actually tested, a `TypeError` for a non-function value, and no special-casing of `unpipe`. Credits the authors of #1307, #789, #796 and #1211 as co-authors.

Closes #1307
Closes #789
Closes #796
Closes #1211
Closes #1189
Closes #572

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ Key | Description
 `limits` | Limits of the uploaded data
 `preservePath` | Keep the full client-supplied path in `file.originalname` instead of just the base name
 `defParamCharset` | Default character set to use for values of part header parameters (e.g. filename) that are not extended parameters (that contain an explicit charset). Default: `'latin1'`
+`streamHandler` | Function that feeds the request body to busboy; defaults to `req.pipe(busboy)`
 
 In an average web app, only `dest` might be required, and configured as shown in
 the following example.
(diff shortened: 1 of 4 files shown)


kilisamemarisaaa (commented):
Reviewed head ddeb6ad and also tested the PR merged synthetically onto the current main, since the branch is several commits behind. On the PR head, the streamHandler/rawBody target tests passed 6/6, the full suite passed 123/123 with 2 pending, and lint passed. On the latest-main merge tree, the same target tests passed 6/6, the full suite passed 138/138 with 2 pending, and lint passed. I found no blocking issues in the stream-handler/raw-body path.

PR #1471, opened 2026-08-29
feat: allow diskStorage without options

vaibhavmashal:
`multer.diskStorage()` without arguments threw `TypeError: Cannot read properties of undefined`, while `multer.memoryStorage()` accepts no arguments. Default `opts` to an empty object so the no-argument form uses the default temporary directory and random filenames, with tests.

Rebased onto `main` as a single commit; the `StorageEngine.md` change was dropped (in that template `destination` returns a full file path, so the string form does not apply there). Changelog entry included.


kilisamemarisaaa (commented):
Could we remove req.file.path before completing the new omitted-options test, and ideally clean up the adjacent default-destination test as well? I ran the new test alone on Windows while snapshotting os.tmpdir(): it passed, but left one new 1,778-byte file whose SHA-256 matched test/files/small0.dat. The full suite leaves two such files because successful uploads do not call _removeFile and the current afterEach only removes uploadDir.

vaibhavmashal:
Thanks for the thorough catch and review @kilisamemarisaaa! 

I've updated both the new omitted-options test and the adjacent default-destination test in \	est/disk-storage.js\ to call \s.unlink(req.file.path, done)\, ensuring any temporary files written to \os.tmpdir()\ are cleaned up before test completion.

Pushed in commit 4ebb5af — all 133 tests continue to pass cleanly.

kilisamemarisaaa (commented):
Reviewed head 4ebb5af. The omitted-options DiskStorage regression is covered, both successful temp-directory tests now unlink req.file.path, and the custom storage template handles omitted opts plus string destinations. I ran npm test (133 passing, 2 pending) and npm run lint locally on Windows. No blocking findings.

vaibhavmashal:
Thank you for the thorough review and verification, @kilisamemarisaaa! Glad everything is clean and passing.

vaibhavmashal:
Just to note regarding the 2 pending tests on Windows: those are the Linux-specific file descriptor tests in \	est/abort-fd-leak.js\ and \	est/abort-unlink-after-close.js\ which check \/proc/self/fd\ and gracefully \skip\ on non-Linux platforms. On Linux (and CI), all 135 tests execute and pass cleanly.

UlisesGascon (approved):

UlisesGascon:
Consolidated the branch: kept your `diskStorage()` fix and tests (rebased onto current `main`, changelog entry added) and dropped the `StorageEngine.md` hunk, since in that template `destination` returns the full file path rather than a directory.

PR #1472, opened 2026-08-30
docs: make StorageEngine template handle string destination

### LINKED ISSUE #1278: "Potential Issue with destination Parameter Type in Multer Custom Storage"

kilisamemarisaaa:
Fixes #1278

## Problem

The custom storage engine template in `StorageEngine.md` assigns
`opts.destination || getDestination` directly to `this.getDestination`, then calls
`this.getDestination(req, file, cb)`. `destination` documented and supported by
`multer.diskStorage()` includes the **string directory path** form — with that form
the template throws `TypeError: this.getDestination is not a function`, because a
string is not callable.

## Fix

- Normalize the option in the constructor exactly the way `storage/disk.js` does:
  when `destination` is a string, wrap it into a `(req, file, cb)` function before
  assigning to `this.getDestination`.
- Add one prose sentence stating that `destination` may be a string path or a
  function, mirroring `multer.diskStorage()`.

Docs-only change; no runtime code touched. The template's behavior now matches
the built-in DiskStorage for both option forms.

## Verification

- Template snippet copy-pasted into a scratch engine and exercised with both
  `destination: '/tmp/uploads'` and `destination: fn` — both resolve a path and
  write the file; the string form no longer throws.
- Docs-only diff (`git diff --stat`: 1 file changed, StorageEngine.md only).

### DIFF
--- a/StorageEngine.md
+++ b/StorageEngine.md
@@ -34,6 +34,10 @@ later on. Multer will decide which files to delete and when. Your storage class 
 implement the `_removeFile` function. It will receive the same arguments as
 `_handleFile`. Invoke the callback once the file has been removed.
 
+The `destination` option may be given either as a string directory path or as a
+`function (req, file, cb)`, mirroring the option accepted by
+`multer.diskStorage()`. The template below accepts both forms.
+
 ## Template
 
 ```javascript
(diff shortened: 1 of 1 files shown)


kilisamemarisaaa:
Closing this in favor of #1471, which was opened earlier and includes the same StorageEngine.md correction together with a tested runtime fix. I also missed the existing draft #1394 in my final duplicate check. Sorry for the duplicate noise; no need to carry a third implementation.

PR #1473, opened 2026-08-31
fix: decode escaped field names, not just filenames

MohammedAlkindi:
`decodeFormDataName` reverses the WHATWG escaping of `%0A`, `%0D` and `%22`, and its own comment says the rule covers "field names and filenames". It was only applied to `filename`. Busboy leaves both escaped, so `req.body` keys, `file.fieldname` and `err.field` all carried the raw escapes.

Measured against the unpatched tree:

```
body keys                        -> ["a%22b"]        expected a"b
upload.single('a"b') + that file -> LIMIT_UNEXPECTED_FILE
originalname                     -> q".txt           already correct
```

So an `<input name='a"b' type=file>` upload is rejected outright, and the only way to accept it was to register the escaped spelling.

`fieldNameSize` now measures the name as it arrived, since decoding shortens it and measuring the decoded form would quietly weaken that limit. `decodeFormDataName` itself is untouched, so the deliberate "never `decodeURIComponent`" property still holds and `50%.pdf` stays intact.

Six new tests, two of which pass before the change as controls (a name with no escapes, and a literal percent sign). The other four fail on the unpatched tree. Each of the two call sites was neutered independently to confirm both are load-bearing.

Suite: 132 passing before, 138 after, 2 pending, no failures either side; `standard` exits 0.

This is the field-name half of #1421, which fixed the filename half.

### DIFF
--- a/lib/make-middleware.js
+++ b/lib/make-middleware.js
@@ -230,12 +230,18 @@ function makeMiddleware (setup) {
     // handle text field data
     busboy.on('field', function (fieldname, value, { nameTruncated, valueTruncated }) {
       if (fieldname == null) return abortWithCode('MISSING_FIELD_NAME')
+
+      // fieldNameSize limits the name as it arrived, so keep the raw form for
+      // that check before the escapes are reversed.
+      var rawFieldname = fieldname
+      fieldname = decodeFormDataName(fieldname)
+
       if (nameTruncated) return abortWithCode('LIMIT_FIELD_KEY')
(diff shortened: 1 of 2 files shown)


kilisamemarisaaa (commented):
Verified on Windows with Node 24.12.0 at 26c5a6b: npm test passes (138 passing, 2 pending) and standard lint is clean. The new field-name cases cover percent-encoding decode, literal percent preservation, expected-file matching, and decoded limit errors; existing filename and storage/cleanup coverage remains green. I found no correctness issues.

UlisesGascon (approved):

Issue #1461, opened 2026-08-28
v3: enforce a default for fieldArrayIndexLimit


UlisesGascon:
`fieldArrayIndexLimit` (added in #1438) is opt-in and defaults to unbounded. On v3, give it an enforced default like #1433 does for `fieldNestingDepth`.

Proposed default: `1000`.

kilisamemarisaaa:
I checked the v3 path and `append-field@2.0.0` to quantify the risk behind an enforced default.

With a numeric bracket path such as `items[1000000]`, `append-field` creates a sparse array whose length is 1,000,001 even though it stores one value. The same parser treats malformed paths such as `items[1000000]tail` as a literal key, so the limit should be applied only to numeric array steps that will actually be materialized.

A focused boundary suite for the proposed default of 1000 could cover:

- `items[1000]` is accepted;
- `items[1001]` is rejected with `LIMIT_FIELD_ARRAY_INDEX` and includes the field name;
- an explicit `limits.fieldArrayIndexLimit: 2000` permits `items[1001]`;
- literal/fallback keys such as `items[1000]tail` are not misclassified as array indices.

This keeps the protection focused on the sparse-array allocation path while preserving an explicit opt-in for applications that need larger indices.

PR #1477, opened 2026-09-02
feat!: enforce default fieldArrayIndexLimit of 1000

official-burak:
Ulises asked for this on v3 in #1461, matching the fieldNestingDepth default in #1433. A name like `items[1000000]` currently builds a huge sparse array even when only one value is sent, so this ports the index check from #1438 and defaults it to 1000.

`items[1000]` still works, `items[1001]` is rejected with `LIMIT_FIELD_ARRAY_INDEX`, and you can raise the limit if you need larger indices. Literal keys such as `items[1000]tail` are left alone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ Key | Description | Default
 `fieldSize` | Max number of bytes per field value | `'8KB'`
 `fields` | Max number of fields per request | `1000`
 `fieldNestingDepth` | Max number of nesting levels for field names (e.g. `a[b][c]` has 2 levels) | Infinity
+`fieldArrayIndexLimit` | Max numeric array index accepted inside a field name (e.g. `a[3]` uses index 3) | `1000`
 `fileSize` | Max number of bytes per file | `'8MB'`
 `files` | Max number of files per request | `10`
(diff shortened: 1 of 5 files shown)


kilisamemarisaaa (commented):
I found a false positive in `exceedsArrayIndexLimit` for a fallback key containing a nested `[`.

`append-field@2.0.0` parses `items[[1001]` as a literal object key (the bracket group is `[1001`), not an array path: its numeric step matcher requires `[` to be followed immediately by digits. However, this PR's global `/\\[(\\d+)\\]/g` scan finds the inner `[1001]` and rejects the field with the default limit:

```text
items[[1001] rejected LIMIT_FIELD_ARRAY_INDEX items[[1001]
items[1001]tail accepted {"items[1001]tail":"value"}
```

I reproduced this on `d6d25b4b` in `multer-pr1477`; the existing 71-test suite and lint still pass. This contradicts the stated “literal fallback keys … left alone” behavior and breaks a previously accepted field name under the new default. Please add a regression test and make the scan follow append-field's actual path grammar (so digits inside an object-key bracket are not treated as an array step).

PR #1483, opened 2026-09-20
docs: clarify preservePath and add #1244 regression tests

### LINKED ISSUE #1244: `preservePath` option not respected
Currently, when passing `preservePath` as an option to multer, it seems that its not respected and the path were not preserved.

The input provided is in the format Buffer, passed like so in the test file using `supertest`:

```typescript
const reqInstance = testApp.post(
 /upload-multiple
);

// Add files
codeFiles.forEach((file) => {
  reqInstance.attach(
    `files`,
    file.data // buffer,
    {filename:  `/some/path/${file.name}.txt` }
  );
});

const response = await reqInstance
  .set("Content-Type", "multipart/form-data")
  .expect(200)
```

The server is setup like so:
```typescript

const app = express();
const upload = multer({ storage: memoryStorage(), preservePath: true });

app.post("/upload-multiple", upload.array("files"), (res) => {
   console.log(res.files[0].originalname) // here it shows only {file.name}.txt, not the full path.
   return "OK"
})
```

Have not tried to replicate this using `busboy` yet, will do so if I manage to find time for the replication.

Is this behaviour expected?

dyk1454683243-sudo:
Fixes #1244

## Summary

`preservePath: true` already forwards the multipart `filename` to busboy and keeps path segments in `file.originalname`. Issue #1244 looked like a multer bug because SuperTest / `form-data` pass the `filename` option through `path.basename` before the body is sent, so a value like `/some/path/a.txt` arrives as `a.txt` and cannot be recovered.

This PR documents that the path must be present on the wire (`filepath` with `form-data`, not `filename`) and adds regression tests for the upload shape from the issue. There is no `lib/` change: inventing server-side path recovery would be incorrect when the client never sent the directories.

## Changes

- README: note that some clients strip directories, and that `form-data` / SuperTest `.attach()` need `filepath` to send path segments.
- `test/preserve-path.js`: hand-built multipart bodies for absolute POSIX and Windows paths, `.array()` + `memoryStorage()`, and `.any()`; plus the original Buffer + `filename` client path to show why that reproduction sees only the base name.

## Test plan

- [x] `npx mocha --reporter spec --exit --check-leaks test/preserve-path.js` — 7 passing
- [x] `npm test` — 172 passing
- [x] `npm run lint` — clean
- [ ] GitHub Actions `ci` / CodeQL: `action_required` (first-time contributor workflow approval)
- [x] Socket Security: pass

Reproduction that matches #1244:

1. Hand-built body with `filename="/some/path/a.txt"` and `preservePath: true` → `originalname === '/some/path/a.txt'`
2. Same body without the option → `originalname === 'a.txt'`
3. `form.append('files', buffer, { filename: '/some/path/a.txt' })` with `preservePath: true` → `originalname === 'a.txt'` (client stripped the path)
4. `form.append(..., { filepath: '/some/path/a.txt' })` → path preserved


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

PR #1484, opened 2026-09-20
test: cover preservePath for absolute and Windows paths

### LINKED ISSUE #1244: `preservePath` option not respected

dyk1454683243-sudo:
Fixes #1244.

## Reproduction

The report used SuperAgent/supertest `.attach({ filename: '/some/path/a.txt' })` with `multer({ storage: memoryStorage(), preservePath: true }).array('files')` and saw `originalname === 'a.txt'`.

That path never reaches Multer. `form-data` (used by SuperAgent) runs the `filename` option through `path.basename` before writing `Content-Disposition`. A hand-built part with `filename="/some/path/a.txt"` already keeps the full path when `preservePath: true`, and strips it when the option is unset — same for `C:\\some\\path\\a.txt`.

`lib/make-middleware.js` already forwards `options.preservePath` to busboy. There is no runtime change.

## Change

- Regression tests for the #1244 path styles through `array()` + `memoryStorage()`, using a raw multipart body so the path is actually on the wire.
- README note that SuperAgent/supertest `.attach({ filename })` strips path segments, and that `form-data`'s `filepath` option (or a raw body) is what `preservePath` can see.

Verified by forcing `preservePath: false` in `lib/make-middleware.js`: the three "keep the path" cases fail (`a.txt` / `c.txt` instead of the full path) and pass again once restored.

`npx mocha --reporter spec --exit --check-leaks test/preserve-path.js` is 6 passing. `npm run lint` is clean.


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

